PR #31, opened 2019-09-10
add gateway service and arbitrary metadata support

labkode:
This PR adds a new service: GatewayService that is supposed to be the only API clients needs to connect to (Phoenix, CLI, ...). The GatewayService acts a as facade for the internal APIS (StorageProvider, StorageRegistry, ....), reducing the complexity of many operations to the end-clients.

The GatewayService is not a superset of the internal APIs like the StorageProvider, it has their own types as it needs to evolve independently from internal APIs, Furthermore, not all available methods on the internal APIs are exposed to the end-clients for simplicity, if we acknowledge that we need to expose more methods, we'll add them step by step.

This PR also includes support for setting/unsetting arbitrary metadata of a resource , for example, favourites, comments, locks, ...

The mechanism to obtain the arbitrary metadata from a resource is to Stat it or ListContainer with the metadata keys wished to be returned, a value of '*' means to return all available arbitrary metadata if possible.


butonic:
Requesting them in the ListCollection rpc feels like a propfind ... and would fit our purposes quite well, obviously. Let me think about additional properties ... eg extended file attributes ... or acls for that matter.

Maybe that is why you separate ArbitraryMetadada from CanonicalMetadata.
Sidenote: besides extended attributes and acls there are even more attributes depenting on the filesystem: https://linux.die.net/man/1/chattr eg to indicate a file is compressed or immutable.

Some namespeces for extended attributes are interesting, eg: `user.dublincore.` or `user.xdg.` which has eg. `user.xdg.comment`, see https://www.freedesktop.org/wiki/CommonExtendedAttributes/

As you already pointed out, for favorites we may need something like `oc.<userid>.favorite=[1|0]`. but there is the issue of having to filter attrubutes which should not be visible to other users ... faverite is one of those attributes.

Considering comments: they should be visible to other users, but not editable. So ... this kind of makes me think it makes more sense to store them in a dedicated location. a files_comments folder per user would be ok to persist the comments. a folder per file id with an id per comment. that way we can even set acls on the comments and allow other users / moderators to change comments ... but it would need an efficient endpoint to retrieve comments. Which brings us to the point where it makes more sense to properly integrate with rocket chat, meet, slack, irc, xmpp or whatever and create a room for every file?

labkode:
@butonic CanonicalMetadata is metadata that is useful for end-users independently of the storage backend underneath  and that is intrinsic to the protocol, mainly for syncing.

Currently there is no namespace only key/value pairs for the ArbitraryMetadata values to start with. I definitely agree that we should come up with a simple namespace convention scoped to a user with certain visibility, from CS3 APIS perspective namespacing is easy as we can have typed strucsts, let's say:

```
message key {
   string namespace = 1;
   string value = 2;
   string user = 3; // probably not needed as we know the user doing the op
}
```

PR #41, opened 2019-11-08
Fix master branch

labkode:


cs3/authregistry/v0alpha/1
```diff
@@ -0,0 +1,101 @@
+// Copyright 2018-2019 CERN
```
tboerger: Added this file by accident?

PR #46, opened 2019-11-21
return user on auth

butonic:
born out of https://github.com/cs3org/reva/pull/382 and the previous discussion in https://github.com/cs3org/reva/pull/239#issuecomment-529524824

The userid is not good enough:
- Depending on the authentication mechanism we can provide more 'claims' than just the id
- splitting auth from userlookup doubles the number of ldap lookups (one to check credentials, one to fetch user metadata ... which could be done in the same request)
- oidc does a user metadata lookup during authentication ... but has no way to forward the resolved claims to the user manager

Even the basic auth provider can return a username and a userid.

The job of the auth provider is to establish the identity of the user and return a stable identifier. But he should be allowed to provide more 'claims'.

I also cleaned up the user type to get rid of the direct `subject`and `issuer` members, which are part of the id. And I added the `mail_verified` member ....

TBH I would just add all oidc standard claims ... in a subsequent PR. For now, this would be awesome!

### DIFF
--- a/cs3/authprovider/v0alpha/authprovider.proto
+++ b/cs3/authprovider/v0alpha/authprovider.proto
@@ -43,8 +43,6 @@ message AuthenticateRequest {
 }
 
 message AuthenticateResponse {
-  reserved 2;
-  reserved "user_id";
   cs3.rpc.Status status = 1;
-  cs3.userproviderv0alpha.User user = 3;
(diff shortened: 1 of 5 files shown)


butonic:
ok, that also updated tha proto.lock ...

butonic:
merged as part of https://github.com/cs3org/cs3apis/pull/48

PR #47, opened 2019-11-25
follow prototool naming practice for 'beta' apis that allow breaking changes

butonic:
@labkode ran into this when digging into prototool and protolock. Maybe you know someone with more experience regarding protobuf. I would follow the documentation from prototool, which seems to document their best practice.

See https://github.com/uber/prototool/blob/dev/docs/breaking.md

Yes, this includes the user pr, but the change is huge and this needs discussion anyway, AFAICT. IMO we should do this change sooner rather than later...


labkode:
@butonic that is really cool, I didn't know prototool had that feature :)

Let me check that protolock may have something also

labkode:
@butonic can you add also the fully qualified name like this:

`
option go_package = "google.golang.org/genproto/googleapis/datastore/v1beta3;datastore";
`

butonic:
AFAICT we need to follow tho prototool way of naming go packages: The [style guide](https://github.com/uber/prototool/blob/dev/style/README.md#package-versioning) says the following:
```go
// Of special note: For the V1 Style Guide, this was the last package sub-name concatenated
// with "pb", so for a package "uber.trip", it would be "trippb", but since the V2 Style Guide
// requires package versions, we have changed "pb" to be the package version.
option go_package = "tripv1";
```

butonic:
merged as part of https://github.com/cs3org/cs3apis/pull/48

PR #49, opened 2019-12-02
Add go-micro generator to prototool.yaml

refs:
Generate go-micro definitions for the cs3 apis. Can this generation be added as an optional `extra_modifier` and end up with a unique file?

- [ ] Add missing dependencies


tboerger:
I like that, this would make our efforts easier

labkode:
@tboerger @refs as we've discussed in the Reva channel I don't mind to have the prototool option there however the generated APIs (Go, Python, ...) won't contain the micro-specfic code. The motivation behind is that if I accept it next time someone comes with another extension point for the APIs used by another framework I'm going to have a hard time to say no based on this precedent.

If you need those generated APIs for ocis, I propose that you build the APIs in your own repo. If there would be enough enough traction from people using the APIs from micro we can re-discuss it, but I prefer to be in the caution zone now, I hope you understand. Thanks!

refs:
closing this :)

PR #56, opened 2020-02-05
Adding child count information to message ResourceInfo

DeepDiver1975:
I'd like to propose an additional optional field for the ResourceInfo message:  child_count

With respect to human facing frontends (like phoenix) it could be valuable information to show the number of contained elements.

Let me know what you think of this. THX

### DIFF
--- a/cs3/storage/provider/v1beta1/resources.proto
+++ b/cs3/storage/provider/v1beta1/resources.proto
@@ -86,6 +86,9 @@ message ResourceInfo {
   // OPTIONAL.
   // Arbitrary metadata attached to a resource.
   ArbitraryMetadata arbitrary_metadata = 14;
+  // OPTIONAL
+  // Number of contained children
+  uint64 child_count = 15;
 }
 
 // CanonicalMetadata contains extra metadata


labkode:
@DeepDiver1975 thanks, it makes sense. 

I just wander if we should introduce some recursive fields like these ones:
https://blog.widodh.nl/2015/04/playing-with-cephfs-recursive-statistics/

Which one could be interested for Phoenix? And also with the child_count did you expect it to be recursive or only for the direct children (1st level)?

DeepDiver1975:
> And also with the child_count did you expect it to be recursive or only for the direct children (1st level)?

Since I know the implications on computing such values I was only thinking about the direct children.

I guess it would be reasonable to add both? Or even split up these values:

- number of files
- number of folders
- number of symlinks
- number of ....
- total number of children

But this might already be much ....

labkode:
@DeepDiver1975 I think it could be doable at least enterprise-grade storage like Ceph and EOS support them out of the box :)

DeepDiver1975:
so we better define a stuct holding all these folder stats?

PR #62, opened 2020-04-03
ocm: Add user to GenerateInviteTokenRequest

ishank011:


labkode:
@ishank011 the token is sent to a mail, not to an user_id. If we knew the user_id we wouldn't have developed the Invitation Flow ;)

ishank011:
@labkode We aren't taking this from the user. The GenerateInviteToken method creates a random token and returns that to the user, which the user can mail to users from other providers.
The workflow is: GET /invite -> GenerateInviteToken -> InviteManager. The manager needs to know with which user the token has to be associated with.

labkode:
@ishank011 clear now :)

labkode:
> @labkode This user_id is to store which user created the token. The GenerateInviteToken method creates a random token and returns that to the user, which the user can mail to users from other providers. The workflow is: GET /invite -> GenerateInviteToken -> InviteManager. The manager needs to know with which user the token has to be associated with.

The user when generating the token will be already logged in to the system, the InviteManager will know what user is asking to generate the token. 
My question is: why you need to expose the user_id into this message? For what I understand only the token will be consumed by higher layers to send the mail.

ishank011:
Yes that is correct. What I was previously thinking is in the HTTP service, we'll get the user from the context and then pass that to the gateway GRPC, but instead we can use the context at the invite manager level only. Thanks!

PR #66, opened 2020-04-20
[WIP] Add list and add ep for ocm core shares

LovisaLugnegard:


ishank011:
These need to be compliant with the definitions at https://rawgit.com/GEANT/OCM-API/v1/docs.html, like the CreateOCMCoreShare method.

labkode:
Hi, I realized that we don't need to add these methods right now as we won't probably need them. This PR, which is the latest state of the OCM implementation by some vendors removes these two endpoints, only relies on the createShare:
https://github.com/cs3org/OCM-API/pull/37

So we keep this PR in our back in case in the future we'll need to add them.

LovisaLugnegard:
I'll close this PR then, doesn't spark joy.

PR #67, opened 2020-04-21
Add Update message to UpdatePublicShareRequest

refs:
As far as I see, currently there is no way to specify which fields on a public share are to be updated. This changes to the API should account for this.

@labkode question regarding differences between link / collaboration. Are updates intended to be atomic by design? as in: one field at a time, or do we want to instead submit an updated version of the share? Atomic updates on a single field will result on a small overhead, either on the program logic or the network.

Also, should we align both messages? collaboration protobuf definition uses a `oneOf` whereas this uses an `enum`, in both cases the end result is the same.


labkode:
Hi @refs, the update currently is atomic and yes, it brings that overhead. The APIs needs to be extended to allow bulk updates. The problem on Go is that if you send many updates in the same struct in Go as has default values you don't know what really was supposed to be updated.

For example if I have:

```
type uploadPayload struct {
  displayName string
  password string
}
```

And I send:
```
up := &{displayName: "Docs"}
```

Does it mean:
* update displayName only (that is the intention)
* update displayName and remove password (that is not the intention)

So we need to define in a field list what are the fields that need to be updated.
Like an array containing the enums or something like that.

If that is urgent, I can take a look.

refs:
Not urgent at all, just wondering. You're right in that an API designed that way is very ambiguous, I'm okay sticking to atomic updates :)

PR #73, opened 2020-05-13
Add optional password  to public link and update docs

labkode:


labkode:
@refs I prefer not to avoid surprises when exposing the public share struct.

PR #76, opened 2020-06-16
Add  Open to gateway API

LovisaLugnegard:
To be able to create an open command in reva.

### DIFF
--- a/cs3/gateway/v1beta1/gateway_api.proto
+++ b/cs3/gateway/v1beta1/gateway_api.proto
@@ -148,12 +148,12 @@ service GatewayAPI {
   rpc UnsetArbitraryMetadata(cs3.storage.provider.v1beta1.UnsetArbitraryMetadataRequest) returns (cs3.storage.provider.v1beta1.UnsetArbitraryMetadataResponse);
   // Creates the home directory for a user.
   rpc CreateHome(cs3.storage.provider.v1beta1.CreateHomeRequest) returns (cs3.storage.provider.v1beta1.CreateHomeResponse);
-  
   // *****************************************************************/
   // ************************ APP PROVIDER ********************/
   // *****************************************************************/
+
   // Returns the iframe url from the WOPI server. The iframe url will let you open the document in the correct online document editor.
-  rpc Open(cs3.app.provider.v1beta1.OpenRequest) returns (cs3.app.provider.v1beta1.OpenResponse);
+  rpc OpenFileInAppProvider(cs3.app.provider.v1beta1.OpenFileInAppProviderRequest) returns (cs3.app.provider.v1beta1.OpenFileInAppProviderResponse);
   // *****************************************************************/
   // ************************ USER SHARE PROVIDER ********************/
   // *****************************************************************/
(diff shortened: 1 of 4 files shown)


glpatcern:
As discussed at the telco, it looks OK to me but I'd like @SamuAlfageme to give his green light as you know better this part.

cs3/gateway/v1beta1/gateway_api.proto
```diff
@@ -308,6 +309,8 @@ service GatewayAPI {
 
   // Creates a new ocm share.
   rpc CreateOCMCoreShare(cs3.ocm.core.v1beta1.CreateOCMCoreShareRequest) returns (cs3.ocm.core.v1beta1.CreateOCMCoreShareResponse);
+  // Returns the iframe url.
+  rpc Open(cs3.app.provider.v1beta1.OpenRequest) returns (cs3.app.provider.v1beta1.OpenResponse);
```
SamuAlfageme, line 313: I would go for a more descriptive name of what the method actually `Open`s  (e.g. `OpenFileInAppProvider()`) and a bit of context on where the iframe URL comes from.

SamuAlfageme: Also, is it a part of the "`OCM CORE`" as the block comment mentions on top? Does it make sense to create a new block for "`APP PROVIDERS`"?

LovisaLugnegard: Good points, will update this!

LovisaLugnegard: Thank you for the comments. I agree that `OpenFileInAppProvider` is a good name! WHat do you think @glpatcern?

glpatcern: Yes! for an API, I like it being verbose.
Also the return URL - being it iframed or not - should reflect the fact that it comes from an external Application. `AppProviderURL`? Any better suggestion?

labkode:
@LovisaLugnegard any update on this one?

cs3/gateway/v1beta1/gateway_api.proto
LovisaLugnegard: 👍 I'll update this

LovisaLugnegard:
@labkode Yes, it's updated now

glpatcern:
(I don't seem to have the ability to "approve" but looks good to me: as discussed, having a more self-explanatory API is excellent here)

labkode:
@ishank011are these changes breaking the recently added provider refactoring?

ishank011:
@labkode no they won't as long as we don't update the go-cs3apis version in reva without adding the corresponding changes. Also, since this is a breaking change, I'd prefer that there's not much of a delay between merging this and the changes in reva, since otherwise anyone wanting to make changes to cs3apis would be blocked.

Also, you might want to update the OpenRequest definition again since I reverted that.

labkode:
@LovisaLugnegard in the Open request the field `  storage.provider.v1beta1.ResourceInfo resource_info = 2;
` should be a reference.

LovisaLugnegard:
@labkode Thanks for finding this, I had forgot to push :)

PR #81, opened 2020-07-23
New CS3 module tx (data transfer).

redblom:
Some issues/questions:

- Protobuf does not like the term 'data' in a namespace, it gives an error: 

```
cs3/datatx/v1beta1/datatx_api.proto:21:1:The name "cs3.datatx.v1beta1" contains the outlawed name "data". Data is a decorator and all types on Protobuf are data, consider merging this information into a higher-level type, or if you must have such a type, Use "Info" instead..
```
I now have named it 'tx' to make it compile. Maybe 'dtx' or simply ‘transfer’ is more appropriate?

- In CreateTransferRequest the resource to be transferred is of type ResourceInfo. If this type is correct then maybe the ResourcePermissions should be expended with a 'transfer' permission.

- I’m not sure if additional fields like eg. ‘opaque’ info are needed in the methods.

- I have chosen to let all 3 methods return a transfer info object (TxInfo).
- These methods also needs to added to the Gateway ?
- What are we going to do with the OCM part?

### DIFF
--- a/cs3/tx/v1beta1/resources.proto
+++ b/cs3/tx/v1beta1/resources.proto
@@ -45,7 +45,7 @@ message TxInfo {
     // The data transfer is accepted by the destination.
     STATUS_TRANSFER_ACCEPTED = 4;
     // The data transfer has started and not yet completed.
-    STATUS_TRANSFER_PENDING = 5;
+    STATUS_TRANSFER_IN_PROGRESS = 5;
     // The data transfer has completed.
     STATUS_TRANSFER_COMPLETE = 6;
     // The data transfer has failed.
--- a/cs3/tx/v1beta1/resources.proto
+++ b/cs3/tx/v1beta1/resources.proto
@@ -28,11 +28,16 @@ option java_package = "com.cs3.tx.v1beta1";
 option objc_class_prefix = "CTX";
 option php_namespace = "Cs3\\Tx\\V1Beta1";
 
+// TxId uniquely identifies a transfer in the transfer provider namespace.
+message TxId {
+  string opaque_id = 1;
+}
+
 // TxInfo represents information about a transfer.
 message TxInfo {
   // REQUIRED.
-  // The transfer id.
-  string id = 1;
+  // The transfer identifier.
+  TxId id = 1;
   // Status represents transfer status.
   enum Status {
     STATUS_INVALID = 0;
--- a/cs3/tx/v1beta1/resources.proto
+++ b/cs3/tx/v1beta1/resources.proto
@@ -34,6 +34,10 @@ import "cs3/types/v1beta1/types.proto";
 
 // TxId uniquely identifies a transfer in the transfer provider namespace.
 message TxId {
+  // REQUIRED.
+  // The internal transfer id used by the service implementor
+  // to uniquely identity the transfer in the internal
+  // implementation of the service.
   string opaque_id = 1;
 }
 
--- a/cs3/tx/v1beta1/resources.proto
+++ b/cs3/tx/v1beta1/resources.proto
@@ -28,6 +28,10 @@ option java_package = "com.cs3.tx.v1beta1";
 option objc_class_prefix = "CTX";
 option php_namespace = "Cs3\\Tx\\V1Beta1";
 
+import "cs3/identity/user/v1beta1/resources.proto";
+import "cs3/storage/provider/v1beta1/resources.proto";
+import "cs3/types/v1beta1/types.proto";
+
 // TxId uniquely identifies a transfer in the transfer provider namespace.
 message TxId {
   string opaque_id = 1;
@@ -38,6 +42,9 @@ message TxInfo {
   // REQUIRED.
   // The transfer identifier.
   TxId id = 1;
+  // REQUIRED.
+  // Reference of the resource to be transfered.
+  cs3.storage.provider.v1beta1.Reference ref = 2;
   // Status represents transfer status.
   enum Status {
     STATUS_INVALID = 0;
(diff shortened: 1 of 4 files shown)


cs3/tx/v1beta1/tx_api.proto
```diff
@@ -0,0 +1,128 @@
...
+  // Requests to cancel a transfer.
+  rpc CancelTransfer(CancelTransferRequest) returns (CancelTransferResponse);
+}
+
+message CreateTransferRequest {
+  // TODO: any additional fields; Opaque information ??
+
+  // REQUIRED.
+  // Identifer of the resource to be transfered.
+  cs3.storage.provider.v1beta1.ResourceId resource_id = 1;
```
ishank011, line 66: This should be Reference as it should be able to create transfers based on just the path as well

cs3/tx/v1beta1/resources.proto
```diff
@@ -0,0 +1,69 @@
...
+option csharp_namespace = "Cs3.Tx.V1Beta1";
+option go_package = "txv1beta1";
+option java_multiple_files = true;
+option java_outer_classname = "ResourcesProto";
+option java_package = "com.cs3.tx.v1beta1";
+option objc_class_prefix = "CTX";
+option php_namespace = "Cs3\\Tx\\V1Beta1";
+
+// TxInfo represents information about a transfer.
+message TxInfo {
```
ishank011, line 32: There's no linking between the transfer ID and the resource. Developers will need to define their own bindings, which should be provided by CS3APIs. Refer the [definition for share](https://cs3org.github.io/cs3apis/#cs3.sharing.collaboration.v1beta1.Share), for example, where we have the ID and resourceID.

ishank011, line 32: Shouldn't we store info about who created the transfer? How do we expect GetTransferStatus to be called? It's unlikely users will remember the transfer ID. Even if they do, they will be able to access transfers created by anybody.

Same for the recipient.

cs3/tx/v1beta1/resources.proto
```diff
@@ -0,0 +1,69 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// In applying this license, CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+syntax = "proto3";
+
+package cs3.tx.v1beta1;
```
ishank011, line 21: Is there a reason for using tx? To a new user, it won't make much sense

cs3/tx/v1beta1/resources.proto
```diff
@@ -0,0 +1,69 @@
...
+option java_outer_classname = "ResourcesProto";
+option java_package = "com.cs3.tx.v1beta1";
+option objc_class_prefix = "CTX";
+option php_namespace = "Cs3\\Tx\\V1Beta1";
+
+// TxInfo represents information about a transfer.
+message TxInfo {
+  // REQUIRED.
+  // The transfer id.
+  string id = 1;
```
ishank011, line 35: IMO, the string id should be encompassed by another object to make it easier to extend in the future, like [ShareID](https://cs3org.github.io/cs3apis/#cs3.sharing.collaboration.v1beta1.ShareId).

ishank011 (changes requested):

cs3/tx/v1beta1/resources.proto
redblom: TX is sort of 'legacy'. The 'original name' was datatx but 'data' is not accepted in the namespace (https://github.com/cs3org/cs3apis/pull/81#issue-455594346). How does 'transfer' sound?

ishank011: Oh okay. I think we can stick with this only then.

cs3/tx/v1beta1/resources.proto
redblom: I believe it's more like: what do we want to expose in TxInfo. The transfer id would be the bare minimum for transfer functionality so the user can do 'status notification' and 'cancel transfer' requests. It's up to the user/EFSS to keep hold of that transfer id if he wants to use these methods.
All necessary info (transfer id, resource, who initiated, destination, etc.) concerning this request must be persisted in the IOP in a transfer db anyway to make data transfer work. This is also where the linking between transfer id and the resource exists (https://github.com/cs3org/cs3apis/pull/81#discussion_r488507054).
But yes maybe it is 'user friendly' to add some additional info in the TxInfo. What do you think would be helpful there ?

cs3/tx/v1beta1/resources.proto
redblom: I created TxId {opaque id}

cs3/tx/v1beta1/resources.proto
```diff
@@ -0,0 +1,69 @@
...
+    // The destination could not be found.
+    STATUS_DESTINATION_NOT_FOUND = 1;
+    // A new data transfer
+    STATUS_TRANSFER_NEW = 2;
+    // The data transfer is awaiting acceptance from the destination
+    STATUS_TRANSFER_AWAITING_ACCEPTANCE = 3;
+    // The data transfer is accepted by the destination.
+    STATUS_TRANSFER_ACCEPTED = 4;
+    // The data transfer has started and not yet completed.
+    STATUS_TRANSFER_PENDING = 5;
```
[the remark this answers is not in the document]
redblom: Done

cs3/tx/v1beta1/tx_api.proto
redblom: Yes

cs3/tx/v1beta1/tx_api.proto
```diff
@@ -0,0 +1,128 @@
...
+  // REQUIRED.
+  // The destination reference the resource is transferred to.
+  // An email address.
+  string dest_email = 2;
```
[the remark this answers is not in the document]
redblom: Yes

cs3/tx/v1beta1/resources.proto
```diff
@@ -0,0 +1,90 @@
...
+option objc_class_prefix = "CTX";
+option php_namespace = "Cs3\\Tx\\V1Beta1";
+
+import "cs3/identity/user/v1beta1/resources.proto";
+import "cs3/storage/provider/v1beta1/resources.proto";
+import "cs3/types/v1beta1/types.proto";
+
+// TxId uniquely identifies a transfer in the transfer provider namespace.
+message TxId {
+  string opaque_id = 1;
```
ishank011, line 37: Add comments for opaque_id

ishank011 (approved):

PR #87, opened 2020-08-14
Extend InitiateFileUploadRequest to support exclusive (atomic) uploads

glpatcern:
This extension is motivated by https://github.com/cs3org/wopiserver/pull/25. Discussing with @ishank011 it looks reasonable to extend the `InitiateFileUploadRequest` to support exclusive operations.

What I propose in this PR is a minimal extension of the protocol, but we should think whether we may need to support other POSIX(-like) flags in the future, and have rather an `enum` instead of just a `bool`. For reference, the years-old `open()` call has plenty of options: https://www.man7.org/linux/man-pages/man2/open.2.html .

### DIFF
--- a/cs3/storage/provider/v1beta1/provider_api.proto
+++ b/cs3/storage/provider/v1beta1/provider_api.proto
@@ -280,6 +280,13 @@ message InitiateFileUploadRequest {
   // REQUIRED.
   // The reference to which the action should be performed.
   Reference ref = 2;
+  // OPTIONAL.
+  // Whether the file is to be uploaded in exclusive mode. Defaults to false.
+  // If true, the request SHALL be processed such that only one of multiple concurrent uploads
+  // to the same target reference MAY succeed, whereas all others MUST fail with CODE_ALREADY_EXISTS.
+  // The semantic is similar to the O_CREAT|O_EXCL POSIX flags.
+  // The request MUST return CODE_NOT_IMPLEMENTED if the provider does not support this mode.
+  bool exclusive = 3;
 }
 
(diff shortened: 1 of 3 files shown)


labkode:
@glpatcern no need to add this as this should be by design, but is not honoured in the implementation (yet).

glpatcern:
Hmm it should still be an option, in the sense that I may want to upload a file "no matter what", overwriting the destination (that's `O_TRUNC` in POSIX) - which is the current behavior. Or I may want to atomically check-and-upload, which is the scope of this new flag.

labkode:
@glpatcern I'm thinking from the perspective of synchronisation, where you create a conflict in case the file has been uploaded in the meantime. What would be the case for "always upload no matter what"

glpatcern:
Good point, but even sync clients would need the other mode to upload an existing file for which they have a legitimate new version. Or do you expect to have the server do delete + upload?

Thinking about other use cases for a "upload no matter what": uploading a file on the web interface and overwrite the destination because the user decided to do so. In general, true that the overwrite mode can be implemented by delete + upload in exclusive mode, but a direct overwrite seems a better option IMHO. At the same time, for the sake of a simpler API it is also reasonable to say "Upload must all be exclusive, no overwrite allowed" and that's it - and make sure the implementations follow.

Also, what do we do with CERNBox today? Do we support only overwrite/truncation, or only exclusive write?

glpatcern:
In fact, as we provide WebDAV support, we probably have no choice but supporting both modes, as we can't rely on the (WebDAV) client to resolve the overwrite into two operations (delete + upload), which on top won't be atomic.

Issue #88, opened 2020-08-21
Add CreateSymlink to storage provider


butonic:
In order to transparently handle symlinks the cs3api should have a dedicated call to create symlinks. We plan to use it so that our settings, accounts and store service can use the cs3 api to persist data.

labkode:
@butonic I need more context on what you're trying to achieve

butonic:
Currently, building a monotonic increasing counter on top of a distributed filesystem using symlinks (or file locks, tracked in https://github.com/cs3org/cs3apis/issues/6).

But we also have in mind being able to sync symlinks with the desktop client. https://github.com/owncloud/client/issues/1440 as have you  https://github.com/owncloud/ocis/issues/43

butonic:
supporting symlinks is not a top priority for the owncloud sync, but we would like to use them with services in ocis. instead of relying on a local storage we would point them to a namespace in the cs3api and could use the storage layer underneath it.

PR #89, opened 2020-08-28
storage provider: add CreateSymlink

butonic:
fixes https://github.com/cs3org/cs3apis/issues/88


IljaN (approved):

IljaN:
@butonic We would also need a function to resolve the symlink to the original filename/path

butonic:
no, the client is responsible for interpreting the link targed. not part of the cs3 api.

ishank011 (approved):

PR #94, opened 2020-10-05
Add FindGroups method to UserProvider API

ishank011:


prototool.yaml
```diff
@@ -30,5 +33,5 @@ generate:
   plugins:
     - name : go
       type: go
-      flags: plugins=grpc
+      flags: plugins=grpc,paths=source_relative
```
labkode, line 36: What is this for?

ishank011: The build https://cloud.drone.io/cs3org/cs3apis/323/1/2 is failing with warnings that the go-package name should be the full path, but prototool doesn't support that. Tried a couple of fixes from here https://github.com/uber/prototool/issues/549 but those aren't working.

I can't seem to figure out where the dependencies got updated. This commit https://github.com/cs3org/cs3apis-build/pull/1 was merged today but that doesn't update anything. Maybe there's an implicit dependency in the dockerfile https://github.com/cs3org/cs3apis-build/blob/master/Dockerfile which got updated when a new image was published.

labkode: The example contains:
      flags: plugins=grpc # didn't appear to need 'paths=source_relative', but outputting to a different location,
@ishank011 can you try without the `paths=source_relative` option?

ishank011: Tried that as well

labkode (approved):

PR #95, opened 2020-11-04
add delete home to the storage provider

C0rby:
Added a delete home method to the storage provider to be able to clean up when removing a user.

### DIFF
--- a/cs3/storage/provider/v1beta1/provider_api.proto
+++ b/cs3/storage/provider/v1beta1/provider_api.proto
@@ -28,6 +28,7 @@ option java_package = "com.cs3.storage.provider.v1beta1";
 option objc_class_prefix = "CSP";
 option php_namespace = "Cs3\\Storage\\Provider\\V1Beta1";
 
+import "cs3/identity/user/v1beta1/resources.proto";
 import "cs3/rpc/v1beta1/status.proto";
 import "cs3/storage/provider/v1beta1/resources.proto";
 import "cs3/types/v1beta1/types.proto";
@@ -144,6 +145,14 @@ service ProviderAPI {
   rpc CreateHome(CreateHomeRequest) returns (CreateHomeResponse);
   // Gets the home path for the user.
   rpc GetHome(GetHomeRequest) returns (GetHomeResponse);
+  // Creates a storage space.
+  rpc CreateStorageSpace(CreateStorageSpaceRequest) returns (CreateStorageSpaceResponse);
+  // Lists storage spaces.
+  rpc ListStorageSpaces(ListStorageSpacesRequest) returns (ListStorageSpacesResponse);
+  // Updates a storage space.
+  rpc UpdateStorageSpace(UpdateStorageSpaceRequest) returns (UpdateStorageSpaceResponse);
+  // Deletes a storage space.
+  rpc DeleteStorageSpace(DeleteStorageSpaceRequest) returns (DeleteStorageSpaceResponse);
 }
 
 message GetHomeRequest {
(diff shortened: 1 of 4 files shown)


cs3/storage/provider/v1beta1/provider_api.proto
```diff
@@ -724,3 +726,18 @@ message CreateHomeResponse {
   // Opaque information.
   cs3.types.v1beta1.Opaque opaque = 2;
 }
+
+message DeleteHomeRequest {
+  // OPTIONAL.
+  // Opaque information.
+  cs3.types.v1beta1.Opaque opaque = 1;
```
refs, line 733: Having `DeleteHomeRequest` as optional leaks implementation details, for instance, we can call `DeleteHome` call with no arguments and this **COULD** translate into "delete home for the authenticated user", having a side effect.

I'm just saying that we have a chance to make this a pure function and have no side effects by adding a [user](https://cs3org.github.io/cs3apis/#cs3.identity.user.v1beta1.User) to the request 🤷

C0rby: I would be fine for that. The reason I did it like that is that I wanted the API to be similar to CreateHome.

butonic: I think CreateHome should also use a user parameter.

labkode:
This API call should be part of an Admin API, we have discussed it in the past and now with this need arising is the right moment to create it. @C0rby can you add this new API here: `cs3/admin/v1beta1/admin_api.proto` with that method?
Note that the method must accept a user ID as argument, this API will be used by site administrators or by robot accounts to automate actions.

C0rby:
Cool.
Just for clarification, should I move the method to the admin_api or copy it to have it in both places?

refs:
> This API call should be part of an Admin API, we have discussed it in the past and now with this need arising is the right moment to create it. @C0rby can you add this new API here: `cs3/admin/v1beta1/admin_api.proto` with that method?
> Note that the method must accept a user ID as argument, this API will be used by site administrators or by robot accounts to automate actions.

thanks for the clarification, I meant `userid`, not an entire user 👯

cs3/admin/v1beta1/admin_api.proto
```diff
@@ -0,0 +1,92 @@
...
+  rpc DeprovisionUser(DeprovisionUserRequest) returns (DeprovisionUserResponse);
+}
+
+message ProvisionUserRequest {
+  // OPTIONAL.
+  // Opaque information.
+  cs3.types.v1beta1.Opaque opaque = 1;
+  // REQUIRED.
+  // The user for which the action should be performed.
+  cs3.identity.user.v1beta1.UserId user_id = 2;
```
labkode, line 64: Let's use cs3.identity.user.v1beta1.User instead to give more information to the system when creating an user.

cs3/storage/provider/v1beta1/provider_api.proto
```diff
@@ -724,3 +729,42 @@ message CreateHomeResponse {
...
+message ProvisionStorageRequest {
+  // OPTIONAL.
+  // Opaque information.
+  cs3.types.v1beta1.Opaque opaque = 1;
+  // REQUIRED.
+  // The owner of the storage.
+  cs3.identity.user.v1beta1.UserId owner = 2;
+  // REQUIRED.
+  // The quota for the user.
+  uint64 quota = 3;
```
labkode, line 742: let's redefine this to:
`uint64 quota_max_bytes`
and add also this field
`uint64 quota_max_files`.

The latter is to limit the number of files a user can have, EOS supports that.

cs3/storage/provider/v1beta1/provider_api.proto
```diff
@@ -724,3 +729,42 @@ message CreateHomeResponse {
...
   cs3.types.v1beta1.Opaque opaque = 2;
 }
+
+message ProvisionStorageRequest {
+  // OPTIONAL.
+  // Opaque information.
+  cs3.types.v1beta1.Opaque opaque = 1;
+  // REQUIRED.
+  // The owner of the storage.
+  cs3.identity.user.v1beta1.UserId owner = 2;
```
labkode, line 739: Let's use a full user object here, like in the Admin API, like that the Storage can have more information (like uid/gid) already provided to provision it without having to call the user service.

This function is called only when provisioning users,  the rate of this call will be very low so we can afford to prefetch this info beforehand.

labkode (changes requested):
@C0rby I've added some comments, overall looks pretty good.

labkode:
@C0rby @refs any update on this? Will be nice to merge it in.

C0rby:
I rebased the branch.
From my POV we can merge, right @butonic?

kulmann:
@C0rby could you rebase please? Would be nice to get this in. :-)

phil-davis:
@C0rby are you online this week? Can you rebase this and we can try to push it along. There are things that depend on this that are waiting.

phil-davis:
@labkode @butonic @refs looks like time to review.

There is no response from drone CI - but that will be related to the drone troubles over Christmas...

C0rby:
Well, new year new luck.

Can we merge this? @refs, @butonic, @labkode

refs (approved):

refs:
@C0rby it LGTM, could we get @ishank011 to have another pair of eyes? :)

ishank011:
@C0rby I think I'm missing something. Why do we have duplicate definitions for (de)provisioning? IMO, the admin API should serve as a gateway to the RPCs in the storage provider, like the gateway API (but we shouldn't add it there so creating a new one makes sense)

ishank011:
@C0rby looks good but how will `ProvisionStorage` be called? I'm guessing from the `ProvisionUser` method in the admin API? How will we supply the quota parameters to it then?

C0rby:
I see your point.
@butonic, what do you say? Add the quota to the ProvisionUser api?

butonic:
As discussed at the CS3 we should pick a more generic name for `CreateHome`. @ishank011 proposed `CreateStorageSpace` which captures the difference between creating a storage vs creating something inside a storage. _Something_ being called a space.

Other options were _CreateTree_, _CreateSpace_, _CreateSubTree_, and _CreateRoot_ but they are either too abstract / graph theory influenced (anything with tree or root) or would invent a new term (space).

_Storage space_ seems to be a good middle ground. Furthermore it can be used not only for user homes but also other _spaces_ eg. project or group _spaces_.

The method needs an owner (which can be a user or a group), optionally quota and opaque data as usual.

cc @ishank011 @C0rby

C0rby:
@jfd, wouldn't CreateStorageSpace also  need a `path` parameter or something like that?

butonic:
CreateStorageSpace where? 

In the gateway? there it would need an indication of which storage should create the space. might be done by path or by storageid.

In a storage provider? No, It is up to the storage provider to determine where to put the new root of the storeage.

In the storage registry? Yes, it needs to know where to mount a storage space. Not only a strorage provitder. But this is a bigger change.

For now, I would limit this change to the StorageProvider Api in the context of the UserProvisioning api: the Gateway can determine the responsible storage provider by calling `storageregistry->GetHome()` and then using a `storageregistry->GetProvider(<home>)`. It can then send the `CreateStorageSpace(<owner>, <quota>)` to it, just as it is currently sending a `CreateHome()`

butonic:
@C0rby  While you are at it could you also add an `UpdateStorageSpace` so we can adjust the quota? We need that for the quota related tests.

butonic:
blocks https://github.com/owncloud/ocis/issues/1290
blocks https://github.com/owncloud/ocis/issues/1313
https://github.com/owncloud/product/issues/247

C0rby:
@ishank011 @labkode, @butonic and I now implemented the storage API how we think it should be.
I removed the admin API again and only added the storage space stuff.

ishank011 (approved):

phil-davis:
@C0rby is there a PR to `cs3org/reva` that implements this?

C0rby:
@phil-davis, not yet. We still have to do that. It will most likely happen within the next sprint.

PR #98, opened 2020-11-25
Add Group API

butonic:
This PR adds a readonly groups api to CS3.
Similar to users groups have a persistent, non reassignable, stable, unique identifier which we call `id`
In addition to that the CS3 API knows other identifiers like a name, an email and numeric identifiers (which we used to carry as opaque payload).

I did not care to preserve backwards compatability and reused old ids, which is why there are build failures for protobuf.

Let me know if we can go forward with this or if the API is considered staple and we need to keep backwards compatabililty.

### DIFF
--- a/cs3/gateway/v1beta1/gateway_api.proto
+++ b/cs3/gateway/v1beta1/gateway_api.proto
@@ -34,6 +34,7 @@ import "cs3/auth/registry/v1beta1/registry_api.proto";
 import "cs3/gateway/v1beta1/resources.proto";
 import "cs3/identity/user/v1beta1/resources.proto";
 import "cs3/identity/user/v1beta1/user_api.proto";
+import "cs3/identity/group/v1beta1/group_api.proto";
 import "cs3/ocm/core/v1beta1/ocm_core_api.proto";
 import "cs3/ocm/invite/v1beta1/invite_api.proto";
 import "cs3/ocm/provider/v1beta1/provider_api.proto";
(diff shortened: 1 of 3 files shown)


cs3/gateway/v1beta1/gateway_api.proto
```diff
@@ -279,8 +280,21 @@ service GatewayAPI {
...
   rpc FindUsers(cs3.identity.user.v1beta1.FindUsersRequest) returns (cs3.identity.user.v1beta1.FindUsersResponse);
+  // *****************************************************************/
+  // ************************ GROUP PROVIDER **************************/
+  // *****************************************************************/
+
+  // Gets the information about a group by the group id.
+  rpc GetGroup(cs3.identity.group.v1beta1.GetGroupRequest) returns (cs3.identity.group.v1beta1.GetGroupResponse);
+  // Gets the information about a group based on a specified claim.
+  rpc GetGroupByClaim(cs3.identity.group.v1beta1.GetGroupByClaimRequest) returns (cs3.identity.group.v1beta1.GetGroupByClaimResponse);
+  // Gets the groups of a group.
```
labkode, line 291: Gets the members of a group

cs3/identity/user/v1beta1/resources.proto
```diff
@@ -45,9 +45,11 @@ message UserId {
 message User {
   UserId id = 1;
   string username = 2;
```
labkode, line 47: Here we can just add the new fields at the end to not limit the breakage.

labkode (changes requested):
@butonic go over my comments and then you can just remove the proto.lock file and run `protolock init`.

You may need to install the protolock tool.

cs3/identity/group/v1beta1/resources.proto
```diff
@@ -0,0 +1,55 @@
...
+  // REQUIRED.
+  // the unique identifier for the group in the scope of
+  // the identity provider.
+  string opaque_id = 2;
+}
+
+// Represents a group of the system.
+message Group {
+  GroupId id = 1;
+  string groupname = 2;
```
ishank011, line 48: group_name

ishank011:
> @butonic go over my comments and then you can just remove the proto.lock file and run `protolock init`.
> 
> You may need to install the protolock tool.

Yep, or you can just run `protolock commit --force` and add the lock file to the commit.

refs:
moved to https://github.com/cs3org/cs3apis/pull/99

PR #99, opened 2020-11-26
Add Group API

refs:
Signed-off-by: Jörn Friedrich Dreyer <jfd@butonic.de>

> This PR adds a readonly groups api to CS3.
Similar to users groups have a persistent, non reassignable, stable, unique identifier which we call id
In addition to that the CS3 API knows other identifiers like a name, an email and numeric identifiers (which we used to carry as opaque payload).

I did not care to preserve backwards compatability and reused old ids, which is why there are build failures for protobuf.

Let me know if we can go forward with this or if the API is considered staple and we need to keep backwards compatabililty.

### DIFF
--- a/cs3/identity/group/v1beta1/resources.proto
+++ b/cs3/identity/group/v1beta1/resources.proto
@@ -25,7 +25,7 @@ option go_package = "groupv1beta1";
 option java_multiple_files = true;
 option java_outer_classname = "ResourcesProto";
 option java_package = "com.cs3.identity.group.v1beta1";
-option objc_class_prefix = "CIU";
+option objc_class_prefix = "CIG";
 option php_namespace = "Cs3\\Identity\\Group\\V1Beta1";
 
 import "cs3/identity/user/v1beta1/resources.proto";
--- a/cs3/identity/group/v1beta1/resources.proto
+++ b/cs3/identity/group/v1beta1/resources.proto
@@ -25,7 +25,7 @@ option go_package = "groupv1beta1";
 option java_multiple_files = true;
 option java_outer_classname = "ResourcesProto";
 option java_package = "com.cs3.identity.group.v1beta1";
-option objc_class_prefix = "CIG";
+option objc_class_prefix = "CIU";
 option php_namespace = "Cs3\\Identity\\Group\\V1Beta1";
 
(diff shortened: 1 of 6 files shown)


refs:
@labkode @ishank011 as I cannot push to `butonic/cs3apis` I needed to fork and apply the review comments on this PR. The other PR can be closed.

cs3/identity/group/v1beta1/resources.proto
```diff
@@ -0,0 +1,55 @@
...
+syntax = "proto3";
+
+package cs3.identity.group.v1beta1;
+
+option csharp_namespace = "Cs3.Identity.Group.V1Beta1";
+option go_package = "groupv1beta1";
+option java_multiple_files = true;
+option java_outer_classname = "ResourcesProto";
+option java_package = "com.cs3.identity.group.v1beta1";
+option objc_class_prefix = "CIU";
```
ishank011, line 28: This needs to be CIG. That's why the build fails

refs: aaaaah, I was wondering, thank you!

ishank011 (approved):

labkode (approved):

ishank011:
@refs @butonic we have cyclic imports (user/resources and group/resources) due to which the build failed https://cloud.drone.io/cs3org/go-cs3apis/115/1/2

refs:
@ishank011 so it seems

https://github.com/cs3org/cs3apis/pull/99/files?file-filters%5B%5D=.proto#diff-cd2eee9df4cadb638fa1535848bc7357f5d18f6d488485924313bda0a2f63584R21 :(

ishank011:
@refs pushing a fix. There's a duplicate method in both the APIs (`user/IsInGroup` and `group/HasMember`). I'll just remove it from the userprovider.

refs:
it should work, but having a go build step for the proto files before it gets to go-cs3api might be a good idea. Not sure if there is a linter for protobuf to catch errors like this, since it is a language specific error, I don't see this happening in JavaScript

ishank011:
Makes a lot of sense. I think it should be pretty simple to add. We'll just need to execute the language-specific actions without pushing to the corresponding repos. I'll create an issue so I don't forget.

PR #100, opened 2020-12-03
Update docs

ishank011:


ishank011:
Merging without review as it only involved running make.

PR #107, opened 2021-02-22
Add GetAcceptedUsers to Invite api

mirekys:
This PR adds a new rpc endpoint to fetch all users that accepted a certain InviteToken with a certain token name/label.

This requires that additional `name` field is also added by this PR to the `InviteToken` resource.
This field holds a user-definable token name, so that the user can later query for all users that accepted the token with this name.

Related [reva#1484](https://github.com/cs3org/reva/issues/1484)

### DIFF
--- a/cs3/ocm/invite/v1beta1/invite_api.proto
+++ b/cs3/ocm/invite/v1beta1/invite_api.proto
@@ -59,6 +59,8 @@ service InviteAPI {
   rpc AcceptInvite(AcceptInviteRequest) returns (AcceptInviteResponse);
   // Retrieves details about a remote user who has accepted an invite to share.
   rpc GetRemoteUser(GetRemoteUserRequest) returns (GetRemoteUserResponse);
+  // Retrieves a list of users that accepted a certain invite token
+  rpc GetAcceptedUsers(GetAcceptedUsersRequest) returns (GetAcceptedUsersResponse);
 }
 
 message GenerateInviteTokenRequest {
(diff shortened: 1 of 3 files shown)


labkode (approved):

ishank011:
@mirekys this doesn't look like a good solution to me. Users shouldn't have to name tokens when generating them. This adds an unnecessary step that affects UX. Instead, I would go for searching over all the users who've accepted a user's invite. There an issue can be that the local user doesn't know about the display names/emails of remote users with which they accepted the invite, but since they're sending the invite, I think we can assume they know their display names anyway. That is how this will work in the UI anyway, the original user will type the remote user's name.

After discussing this with @labkode, I propose removing `name` from the token and adding a method `FindAcceptedUsers`, similar to the [user API](https://github.com/cs3org/cs3apis/blob/master/cs3/identity/user/v1beta1/user_api.proto#L62), which will search over the accepted users' attributes (display name, email, etc.). We can add one optional field to the invite token `description` in which the original user can add some notes to be sent to the remote users in the mail.

milkydan:
@ishank011 we have been thinking about such a use case. In case you have some Slack channel you wish to create the group from the users contained in that channel. So you'll generate an invitation (QR code) and you will insert it into the Slack channel. Then you don't have any idea who accepted the invitation (you don't know the email and so on), so you cannot find the particular user. That is the reason why we needed an optional field in the invitation so we can easily find particular users eg. from the Slack channel over accepted users.

ishank011:
@milkydan this isn't something I'd worry about.
- If we compare this to the sharing with the local users, you do need to know some attribute about the user you want to share with. So that is a prerequisite that we can assume on the users' side.
- Since the field is optional, if the user doesn't enter a name, there's no way to search for who accepted it.
- Also it won't scale. If a user generates 10 tokens with different names, expecting him to recall these names to find users is not ideal.

We haven't enabled group sharing yet, that is something that is on the agenda for this year and needs to be discussed. This can be one way which we can consider but for the time being, for individual user sharing, I think searching over user attributes would be the better method.

milkydan:
@ishank011 ok we discussed it and agreed with your proposal. Thx :+1:

PR #109, opened 2021-03-03
Share type added to specify Share purpose.

redblom:
It should be possible to persist the purpose of the share together with the share itself, so the application can act accordingly when eg. a received share is accepted. For this the `share_type` field is defined. 
Currently `SHARE` and `TRANSFER` share types are acknowledged.
See https://github.com/cs3org/reva/pull/1242#issuecomment-788932923.

### DIFF
--- a/cs3/sharing/ocm/v1beta1/resources.proto
+++ b/cs3/sharing/ocm/v1beta1/resources.proto
@@ -72,10 +72,11 @@ message Share {
   cs3.types.v1beta1.Timestamp mtime = 9;
   // Share type defines purpose.
   enum ShareType {
+    SHARE_TYPE_INVALID = 0;
     // A regular file or folder share (this is the default share type).
-    SHARE = 0;
+    SHARE_TYPE_REGULAR = 1;
     // A file or folder transfer.
-    TRANSFER = 1;
+    SHARE_TYPE_TRANSFER = 2;
   }
   // Specifies the type of the share so  
   // the service implementor can act accordingly.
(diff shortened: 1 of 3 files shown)


redblom:
The enforced default share type enum value `SHARE_TYPE_INVALID` is ugly.

ishank011:
Please run `make` before pushing

ishank011:
We need to make sure to specify `SHARE_TYPE_REGULAR` at all places in the existing code.

cs3/sharing/ocm/v1beta1/resources.proto
```diff
@@ -70,6 +70,17 @@ message Share {
...
+  // Share type defines purpose.
+  enum ShareType {
+    SHARE_TYPE_INVALID = 0;
+    // A regular file or folder share.
+    SHARE_TYPE_REGULAR = 1;
+    // A file or folder transfer.
+    SHARE_TYPE_TRANSFER = 2;
+  }
+  // Specifies the type of the share so
+  // the service implementor can act accordingly.
```
ishank011, line 82: // Specifies the type of the share.

cs3/sharing/ocm/v1beta1/resources.proto
```diff
@@ -70,6 +70,17 @@ message Share {
   // REQUIRED.
   // Last modification time of the share.
   cs3.types.v1beta1.Timestamp mtime = 9;
+  // Share type defines purpose.
```
ishank011, line 73: // Defines the type of share based on its origin.

ishank011 (approved):

Issue #110, opened 2021-03-05
Authentication methods for password protected link shares.


C0rby:
## tl;dr
For ownCloud Web we needed a different way to authenticate downloads of password protected link shares. The current design of the share manager doesn't allow us to implement this alternative authentication method namely pre-signed urls.
We are using the share password hash as the signing key but in the current implementation the share password hash is not accessible outside of the share manager.

To allow this alternative authentication method I propose three options:
- Include the password hash in the `PublicShare` struct
  - The simplest solution
  - May leak the password hash when the PublicShare struct is included in a response
- Add the password hash as a return value of `GetPublicShare` and `GetPublicShareByToken`
  - Second simplest solution
  - Doesn't leak the hash if the `PublicShare` is included in a response
- Implement different authentication strategies which can be used in the share manager
  - Complex to implement
  - Mixes transport specific auth stuff with the share manager 
  

## Long explanation

This `http://localhost:8080/index.php/s/2LfFihmw6cD7qxU` is an example public link. The last part `2LfFihmw6cD7qxU` is the random token identifying the share.
For password protected shares the credentials can be included in the 'Authorization' header using basic auth where 'public' is the username and the share password is the password.

When we want to download a file using the anchor tag `<a href="...."` we can't supply the Authorization header to the request. 
In ownCloud 10 this worked by using cookies but in ownCloud Web we don't have cookies. 
Because of that we have decided to use signed URLs. The advantage of signed-URLs is that, just like with unprotected public links, the possession of the URL is proof of authorization. Meaning anyone who has the signed URL can download the resource.

The general idea is this:

    The client sends a PROPFIND request to the backend (including the credentials in the auth header)
        The client has to request the `downloadURL` attribute to receive the signed URL
    The backend does the normal credential check
    The backend generates the PROPFIND response
        When the client requested the `downloadURL` attribute the signed URL is generated and returned as the `downloadURL`
    The client can now issue GET requests the `downloadURL`

The detailed signing process looks as follows:

Let's assume we have a shared tree like this:

    Share Root
        some-file.txt
        Folder/
            another-file.txt

To create the signature we need:

    The resource path relative to the share root: `/Folder/another-file.txt`
    A timestamp of the expiration date for the signature. Right now it's 30 minutes from the moment of signing
    The share token
    The share password (hash)

And then we calculate the HMAC like this:

`HMAC_SHA512/256(SHA256(share_password_hash), share_token + resource_path + timestamp)`

[...]

PR #113, opened 2021-03-12
get public share methods also return the share password now

C0rby:
For some alternative public share authentication methods we need to be able to access the public share password. Some more details are documented here: https://github.com/cs3org/cs3apis/issues/110


labkode:
@C0rby let's call the field `password_hash` instead, I think it brings more semantic to someone implementing the API.
The rest looks good.

C0rby:
@labkode, I renamed the field and rebased the PR so now it should be mergable.

labkode (approved):

PR #114, opened 2021-03-12
add etag to versions

wkloucek:
FileVersion must have an Etag property so that oCIS can provide them on the file versions endpoint (https://doc.owncloud.com/server/developer_manual/webdav_api/files_versions.html#file-is-found-and-has-multiple-versions).

Implementation WIP in Reva: https://github.com/cs3org/reva/pull/1526/commits/f2cadc43cf47c090dccbe35399393d78f19937f2

### DIFF
--- a/cs3/storage/provider/v1beta1/resources.proto
+++ b/cs3/storage/provider/v1beta1/resources.proto
@@ -294,6 +294,10 @@ message FileVersion {
   // REQUIRED.
   // The Unix Epoch timestamp in seconds.
   uint64 mtime = 5;
+  // REQUIRED.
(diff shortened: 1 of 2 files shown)


butonic:
@labkode can you get the etage of a version from eos?

labkode:
@butonic the versions are frozen by design. A version must not be updated (that's the idea for a version). So given that you have an unique ID for a file version as per CS3API, you can generate Etag from this behaviour. This should work for any storage.

ishank011:
@butonic @labkode eos does in fact return the etag for each version, which we ignore in eosfs at the moment.

ishank011 (approved):

PR #118, opened 2021-03-23
Support signature authentication for public links

C0rby:
Deprecated the changes from https://github.com/cs3org/cs3apis/pull/113
and enhanced the api to support signature based authentication for
public links.

### DIFF
--- a/cs3/sharing/link/v1beta1/link_api.proto
+++ b/cs3/sharing/link/v1beta1/link_api.proto
@@ -173,6 +173,9 @@ message ListPublicSharesRequest {
   // OPTIONAL.
   // The list of filters to apply if any.
   repeated Filter filters = 2;
+  // OPTIONAL.
+  // If a signature should be included in the share.
+  bool sign = 3;
 }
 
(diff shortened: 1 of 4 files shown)


ishank011:
The signature should ideally be returned when calling `ListPublicShares` as well. Would it make more sense to add it to the `PublicShare` message?

C0rby:
> The signature should ideally be returned when calling `ListPublicShares` as well. Would it make more sense to add it to the `PublicShare` message?

True if we want to introduce this signature authentication in general. I had just had the owncloud public shares in mind which only need the signatures after calling `GetPublicShareByToken` but I can add that to `ListPublicShares`.

C0rby:
Should we always include the signature or should it be controllable via a flag?

ishank011:
I would prefer that the sign be a part of the `PublicShare` message. The flag looks fine. Can you just remove the deprecated field altogether? We don't have a lot of clients so breaking changes should be fine at this stage

ishank011 (approved):

PR #119, opened 2021-04-14
Write only management API for User and Group resources

refs:
## What?

Essentially a write-only management API for User and Group resources. The CS3 API already provides the means for retrieving information for resources such as users and groups but it does not dictate how such resources are created. These minimal changes would add a different write layer to an administration API that would deal with destructive actions. Ultimately admin actions MUST only be taken by users with the role of administrator.

Adding roles will not be yet accounted for in this patch, and implementations should instead rely on their own logic to deal with roles, but a PR will follow up that would add roles to the CS3 API.

## Design Choices

- Namespace (separate) users from groups write/delete operations instead of delegating all to a single Admin API. I considered this as an approach taken in places such as identity.
- Write operations SHOULD always return the latest version of a resource, therefore Write methods are required to return a whole Group and User resource.
- No special status codes needed to be added considered the minimal additions.


ishank011:
@refs can you run `make` and push? That'll generate the documentation

cs3/admin/group/v1beta1/group_api.proto
```diff
@@ -0,0 +1,127 @@
...
+  rpc RemoveUser(RemoveUserRequest) returns (RemoveUserResponse);
+}
+
+message CreateRequest {
+  // OPTIONAL.
+  // Opaque information. Allow to send any arbitrary data a service might use that is outside the API boundaries
+  cs3.types.v1beta1.Opaque opaque = 1;
+  // REQUIRED.
+  // The information of group to be created.
+  cs3.identity.group.v1beta1.Group group = 3;
```
ishank011, line 54: Field number should be 2

cs3/admin/group/v1beta1/group_api.proto
```diff
@@ -0,0 +1,127 @@
...
+}
+
+message DeleteRequest {
+  // OPTIONAL.
+  // Opaque information. Allow to send any arbitrary data a service might use that is outside the API boundaries.
+  cs3.types.v1beta1.Opaque opaque = 1;
+  // REQUIRED.
+  // The group to be deleted, given their ID.
+  cs3.identity.group.v1beta1.GroupId groupid = 2;
```
ishank011, line 75: Change this to `group_id`? Ditto for all other occurrences.

cs3/admin/user/v1beta1/user_api.proto
```diff
@@ -0,0 +1,80 @@
...
+  rpc Delete(DeleteRequest) returns (DeleteResponse);
+}
+
+message CreateRequest {
+  // OPTIONAL.
+  // Opaque information. Allow to send any arbitrary data a service might use that is outside the API boundaries
+  cs3.types.v1beta1.Opaque opaque = 1;
+  // REQUIRED.
+  // The information of user to be created.
+  cs3.identity.user.v1beta1.User user = 3;
```
ishank011, line 49: Field number should be 2

refs:
thanks for the review! good catch with the offset attribute ^^

cs3/admin/group/v1beta1/group_api.proto
```diff
@@ -0,0 +1,127 @@
...
+  cs3.rpc.v1beta1.Status status = 1;
+  // OPTIONAL.
+  // Opaque information.
+  cs3.types.v1beta1.Opaque opaque = 2;
+}
+
+message AddUserRequest {
+  // REQUIRED.
+  // ID of the user to add to the group
+  cs3.identity.user.v1beta1.UserId userid = 1;
```
ishank011, line 90: `user_id` as well here? 😅

ishank011 (approved):
LGTM, but would more descriptive method names be helpful? Eg, `CreateGroup`, `AddUserToGroup`?

refs:
Oh, I was thinking the same, then did a little read on [Google's guidelines](https://cloud.google.com/apis/design/standard_methods) and left the namespace be more explicit. BUT apparently I did not have a look at the examples which explicitly append the resource name to the method haha, let me amend the changes

refs:
@ishank011 just renamed the methods to stick to these guidelines 👍

ishank011 (approved):

Issue #106, opened 2021-02-04
Extend UserID/GroupID definition with type support


labkode:
An account management system, like LDAP, can contain different type of accounts with different connotations.

For example, a primary, secondary, service, or guest account.

I suggest to add an free text format field for time being.

PR #120, opened 2021-04-21
Add user type field to user ID

ishank011:
Need to review if the list of account types is exhaustive.

Closes https://github.com/cs3org/cs3apis/issues/106


refs:
@ishank011 just a question regarding wording and naming conventions, should we refer as this as `UserAccount`, or simply as `User` or `Account`? There is no mention of the concept of `Account` anywhere in the CS3 API since it only deals with Users.

I'm just bringing this up because I recently opened https://github.com/cs3org/cs3apis/pull/119 and had the same internal struggles 🤷‍♂️

ishank011:
@refs ah yes, makes sense. I'll remove the term 'account'. Thanks!

refs (approved):

labkode (approved):

PR #121, opened 2021-04-22
Added a new OpenInApp request/response

glpatcern:
This will eventually supersede `OpenFileInAppProvider`.

The rationale being that `OpenInApp` generically refers to `ref`s, i.e. files AND directories, and a new optional parameter `app` was added, to allow:
1) In case of folders, to specify the app as the default logic to map mimetypes to apps cannot apply
2) In case of files, to allow overriding the default logic and support the case where the user asks to "Open with..." and multiple applications are registered for the same mimetype (typical case: `.docx` -> MS Office Online, OnlyOffice, Collabora Online).

### DIFF
--- a/cs3/app/registry/v1beta1/registry_api.proto
+++ b/cs3/app/registry/v1beta1/registry_api.proto
@@ -82,8 +82,24 @@ message ListAppProvidersRequest {
   // OPTIONAL.
   // Opaque information.
   cs3.types.v1beta1.Opaque opaque = 1;
-  // TODO(labkode): maybe add some filter?
-
+  // REQUIRED.
+  // Represents a filter to apply to the request.
+  message Filter {
+    // The filter to apply.
+    enum Type {
+      TYPE_INVALID = 0;
+      TYPE_MIME_TYPE = 1;
+    }
+    // REQUIRED.
+    Type type = 2;
+    oneof term {
+      // if present, the response MUST list all known app providers for the given mimetype.
+      string mime_type = 3;
+    }
+  }
+  // OPTIONAL.
+  // The list of filters to apply if any.
+  repeated Filter filters = 2;
 }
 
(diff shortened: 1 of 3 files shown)


SamuAlfageme (approved):
LGTM! All in for this open request version in steroids! 🙌

ishank011 (commented):
Maybe just add the optional `app` field to the current request? If it's empty, do what you're doing now, else the changed implementation

cs3/app/registry/v1beta1/registry_api.proto
```diff
@@ -82,8 +82,9 @@ message ListAppProvidersRequest {
   // OPTIONAL.
   // Opaque information.
   cs3.types.v1beta1.Opaque opaque = 1;
-  // TODO(labkode): maybe add some filter?
-
+  // OPTIONAL.
+  // Return all known app providers for a given mimetype (default to all).
+  string mime_type = 2;
```
ishank011, line 87: Maybe use filters? 

https://github.com/cs3org/cs3apis/blob/4b6237bcf90e4eb7d5d12782a2c7f2292ffb401e/cs3/sharing/collaboration/v1beta1/collaboration_api.proto#L145-L164

glpatcern: Ah, that's what is meant by `Filters`, I'll have a look - and sure, the semantic is that: being able to query the app registry by mimetype

glpatcern:
> 
> 
> Maybe just add the optional `app` field to the current request? If it's empty, do what you're doing now, else the changed implementation

Not sure I understand - this is precisely what this PR is doing. Or do you mean to modify also the `OpenFileInAppProvider` request? But the idea is to deprecate it - essentially have it renamed to `OpenInApp`, without references to File.

labkode (approved):

PR #122, opened 2021-04-22
Extend InitiateFileUpload to support exclusive uploads and etags-based checks

glpatcern:
This is a revamp of https://github.com/cs3org/cs3apis/pull/87, which got closed because the `master` branch got renamed to `main`...

Pasting here the relevant parts:

This extension is motivated by cs3org/wopiserver#25. Discussing with @ishank011 it looks reasonable to extend the `InitiateFileUploadRequest` to support exclusive operations: I may want to upload a file "no matter what", overwriting the destination - which is the current behavior. Or I may want to atomically check-and-upload, which is the scope of this new flag. Both actions are legitimate and should be supported.

The PR proposes a minimal extension of the protocol. May we need to support other POSIX(-like) flags and have rather an enum instead of just a bool? I guess by now we're settled, yet for reference the years-old `open()` call has [plenty of options](https://www.man7.org/linux/man-pages/man2/open.2.html).

Related to this, there's been quite some discussion around locking within ownCloud, that's why I'm involving @butonic too. 

Comments/opinions welcome of course, but I'd like to move forward so to start testing the WOPI server locking logic with oCIS.

### DIFF
--- a/cs3/storage/provider/v1beta1/provider_api.proto
+++ b/cs3/storage/provider/v1beta1/provider_api.proto
@@ -280,6 +280,13 @@ message InitiateFileUploadRequest {
   // REQUIRED.
   // The reference to which the action should be performed.
   Reference ref = 2;
+  // OPTIONAL.
+  // Whether the file is to be uploaded in exclusive mode. Defaults to false.
+  // If true, the request SHALL be processed such that only one of multiple concurrent uploads
+  // to the same target reference MAY succeed, whereas all others MUST fail with CODE_ALREADY_EXISTS.
+  // The semantic is similar to the O_CREAT|O_EXCL POSIX flags.
+  // The request MUST return CODE_NOT_IMPLEMENTED if the provider does not support this mode.
+  bool exclusive = 3;
 }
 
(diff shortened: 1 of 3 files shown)


glpatcern:
Following discussions, a quick summary. We should follow the `if-match` etags preconditions used in HTTP headers and implement:
* A more explicit `upload-if-absent` flag (rename `exclusive`)
* A `upload-if-match` option to handle the etag use-case from sync clients

We agree that the default behavior without options is to overwrite the destination without further checks. Name of the optional arguments to be agreed.

Also, the response MUST contain the current etag when `upload-if-match` is used and the precondition fails.

glpatcern (commented):
Reviewing myself, two comments:
1) Shall we use a `oneof` grup for `if_not_exist` and `if_match`, i.e. make them alternative? It does not make sense to pass both of them.
2) In case of etag mismatch, do we want the current etag to be returned as `opaque` or do we give a specific (again optional) field in the `InitiateFileUploadResponse`?

ishank011 (approved):

labkode (approved):

PR #123, opened 2021-04-27
Add token scope field to gateway authenticate requests

ishank011:
We mint the JWT tokens in Gateway/Authenticate. Adding an optional `token_scope` field would allow us to ensure that access is restricted to only these scopes. To be used in cases when we impersonate other users, including public shares, OCM shares and guest accounts.

### DIFF
--- a/cs3/auth/provider/v1beta1/resources.proto
+++ b/cs3/auth/provider/v1beta1/resources.proto
@@ -1,0 +1,62 @@
+// Copyright 2018-2019 CERN
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// In applying this license, CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+syntax = "proto3";
+
+package cs3.auth.provider.v1beta1;
+
+option csharp_namespace = "Cs3.Auth.Provider.V1Beta1";
+option go_package = "providerv1beta1";
+option java_multiple_files = true;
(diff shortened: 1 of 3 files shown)


refs:
This API makes complete sense within the file sync and shares context, which is the scope of the CS3 APIs; however 🤔 I recently started an issue about dealing with roles in a decentralized manner, and was wondering how would this API design choice affect it. On my issue I mention let the IDP define the roles, but IDP's are not subjected to the CS3 APIs.

Are this changes the predecessor of a sort of [Roles Service?](https://github.com/cs3org/cs3apis/issues/26)

ishank011:
Hi @refs. Yes, going by https://github.com/owncloud/ocis/issues/1971, it serves a similar purpose. Primarily, it is used to mint tokens specifically for cases where we impersonate other users without allowing the full scope of what the actual user has access to.

We can discuss how to incorporate both the things in a single design. I wanted to finish coding it so that I could verify it's correctness. We can change the design according to what needs we want it to suffice. Here's a diagram detailing how the tokens are minted and additional scopes added https://codimd.web.cern.ch/XTib-1TzTyqx2IZJJOq5pA (you might have to zoom in a bit)

ishank011:
Also, this serves as a better approach compared to https://github.com/cs3org/reva/issues/1549. Thanks to @labkode for the idea!

cs3/auth/provider/v1beta1/resources.proto
```diff
@@ -0,0 +1,62 @@
...
+option php_namespace = "Cs3\\Auth\\Provider\\V1Beta1";
+
+import "cs3/types/v1beta1/types.proto";
+
+// Scope defines role-based permissions for various resources.
+message Scope {
+  // REQUIRED.
+  // The resource embedded in the request of a particular method. It depends on
+  // the method, hence is left as opaque.
+  cs3.types.v1beta1.OpaqueEntry resource = 1;
```
ishank011, line 38: Note opaque here is an opaque object, not a storage reference. That's because we need to allow a variety of requests, not just limited to the storage provider. Eg, `GetPublicShare ref:<token:"XRKWRTefdZPPVVo">`

refs: makes sense, a valid concern. Thanks for pointing it out 👍

refs (approved):
Changes make sense, I think this is good to go. The only piece of criticism I find is the definition of roles at the API level, since they are almost correlated to a "files" app. But as a way to verify a proof of correctness is good to go! 🚀

labkode (approved):

PR #126, opened 2021-05-17
radically simplify references

butonic:
more radical approach than https://github.com/cs3org/cs3apis/pull/125


refs:
@butonic I feel that https://github.com/cs3org/cs3apis/pull/126/commits/db8f74fff4f7f08e87eb8950e0d838b784ccc7de made things clearer, and as per understanding, the language we use easier. Relying in the combination of storage_id + node_id + path should be enough to identify any resource using references alone 👍

butonic:
superseded by https://github.com/cs3org/cs3apis/pull/130

PR #130, opened 2021-06-07
Simplify references for retrieval

butonic:
cleaned up version of https://github.com/cs3org/cs3apis/pull/126 as a new PR to have them comparable.

- simplifies the references into a combined `Reference` with `storage_id`, `node_id` and `path`
- leaves property names for id based references as `resource_id` but does change the type. The share implementations need to use ids to prevent breaking references when a path changes. 
- leaves the type `string` `Path` property in `ResourceInfo`

### DIFF
--- a/cs3/storage/provider/v1beta1/resources.proto
+++ b/cs3/storage/provider/v1beta1/resources.proto
@@ -189,17 +189,17 @@ message ResourceChecksumPriority {
 }
 
 // The mechanism to identify a resource in the CS3 namespace.
-// It can represent path based references as well as combined references:
-// The storage registry uses the storage id to determine the responsible storage provider. When the storage space id is not available it will use the path.
+// It can represent path based, id based and combined references:
+// The storage registry uses the storage_id to determine the responsible storage provider.
+// When the storage_id is not available it will use the path.
 // In a URL the different components can be represented in a string using the following layout:
 // <storage_id>!<node_id>:<path>
 message Reference {
   // OPTIONAL.
-  // The id of the storage. Used by the storage registry to determine the responsible storage provider.
+  // The logical id of a storage. Used by the storage registry to determine the responsible storage provider.
   string storage_id = 1;
   // OPTIONAL.
-  // The node id used by a storage provider to uniquely identity a resource in the internal
-  // implementation of the service.
+  // The logical node id used by a storage provider to uniquely identify a resource in a storage.
   string node_id = 2;
   // OPTIONAL.
   // When starting with `/` the reference represents an absolute path. In this case the storage_id and the node_id must be empty.
--- a/cs3/storage/provider/v1beta1/resources.proto
+++ b/cs3/storage/provider/v1beta1/resources.proto
@@ -44,7 +44,7 @@ message ResourceInfo {
   ResourceType type = 2;
   // REQUIRED.
   // Opaque unique identifier of the resource.
-  ResourceId id = 3;
+  Reference id = 3;
   // REQUIRED.
   // The data checksum for the file resource.
   // For all other resources, the checksum is unset.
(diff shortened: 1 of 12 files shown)


cs3/storage/provider/v1beta1/resources.proto
```diff
@@ -313,9 +316,8 @@ message RecycleItem {
   // The key to identify the deleted resource.
   string key = 3;
   // REQUIRED.
-  // The original path of the deleted resource.
-  // MUST start with the slash character (/).
-  string path = 4;
+  // The original reference of the deleted resource.
+  Reference ref = 4;
```
ishank011, line 320: @butonic this change might lead to some inconsistencies. Do we want to return either of resource ID and path here? Won't having both `ResourceId` and `path` fields like in `ResourceInfo` make more sense?

cs3/storage/provider/v1beta1/resources.proto
```diff
@@ -189,27 +190,29 @@ message ResourceChecksumPriority {
...
-  oneof spec {
-    // The path to the resource.
-    // MUST start with the slash character (/).
-    string path = 1;
-    // The id for the resource.
-    // MUST NOT start with the slash character (/).
-    ResourceId id = 2;
-  }
+  // OPTIONAL
+  ResourceId resource_id = 1;
```
refs, line 201: There is one point I don't understand here. From an API perspective this is a bit inconsistent. If both properties are optional, then an empty Reference should be valid, by definition, but an empty reference is not helpful at all. In the previous design this was ensured not to happen because it was defined as a `oneof` relationship:

> If you have a message with many fields and where at most one field will be set at the same time, you can enforce this behavior and save memory by using the oneof feature.

I would update the docs to reflect that one MUST be present.

refs (approved):

cs3/storage/provider/v1beta1/resources.proto
ishank011: +1

cs3/storage/provider/v1beta1/resources.proto
butonic: Restoring trash is similar to creating directories or uploading files. You actually need to pass a reference for the parent and a name for the file. When listing a directory you get `ResourceInfo` items, which carry the `id` of the resource and a path.

When listing `RecycleItem`s the `key` is what you need to use when making a `RestoreRecycleItem` request. So that is the crucial bit of information. I don't think we need an `id`. If we add it clients might want to stat a file in the trash. While I like the idea of being able to navigate a trashed folder, eg. by using ListContainer with that trashed id ... it might not be possible to implement on every storage driver. Should the CS3 API allow it? I think so. What is your take on this?

But when listing `RecycleItem`s there is no relative anchor. Currently, it is not even possible to limit the listing to a certain storage space.

- [ ] Add `ref:Reference` to `ListRecycleRequest`

With that in place we can apply the same logic to the `RecycleItem` `path` property as we have it for `ResourceInfo`:
- when no `Ref.ResouceID` was given in the request the `RecycleItem` `path` property in the response is absolute and must start with `/`
- when a `Ref.ResouceID` was given in the request the `RecycleItem` `path` property in the response is relative and must start with a `.`

cs3/storage/provider/v1beta1/resources.proto
```diff
@@ -64,8 +64,9 @@ message ResourceInfo {
   // For reference types this is NOT the mtime of the target.
   cs3.types.v1beta1.Timestamp mtime = 7;
   // REQUIRED.
-  // The path for the resource.
-  // It MUST start with the slash character (/).
+  // The path for the resource:
+  // MUST start with `/` when the reference had no resource_id, indicating an absolute path.
+  // MUST start with `.` when the reference had a resource_id, indicating a relative path.
```
labkode, line 69: MUST be `` when reference is an ID and returned information is a full path.

butonic: When you want to look up a path, the Reference used in the stat request MUST have a `reference_id` set and the `path` MUST be empty. But the `ResourceInfo` then MUST return the absolute path, which MUST start with `/`.

```
// MUST start with `.` when the reference had a resource_id and a path starting with `.`, indicating a relative path.
// Otherwise MUST start with `/` indicating an absolute path. 
```

I still don't like the naming of references: either they are absolute, or they are relative to the resource id. If we can agree on `relative reference` (id set & path starts with `.`) and `absolute reference` (id empty and path starts with `/`) I propose:

```
// MUST start with `/` when the call used an absolute reference (path starts with `/` & id not set).
// MUST start with `/` when the call used an relative reference (path starts with `.` & id set).
```

cs3/storage/provider/v1beta1/resources.proto
ishank011: > With that in place we can apply the same logic to the RecycleItem path property as we have it for ResourceInfo

LGTM!

> While I like the idea of being able to navigate a trashed folder, eg. by using ListContainer with that trashed id ... it might not be possible to implement on every storage driver. Should the CS3 API allow it? I think so. What is your take on this?

According to me, this can be a pretty nice addition. Of course, some drivers might not implement it which should also be fine. `gateway. ListRecycleRequest` already has a ref parameter. We can use the same request in `storageprovider` but we can extend it later on.

Can you update the comment to reflect @refs's point? I'll merge then

cs3/storage/provider/v1beta1/resources.proto
ishank011: > I still don't like the naming of references: either they are absolute, or they are relative to the resource id. If we can agree on relative reference (id set & path starts with .) and absolute reference (id empty and path starts with /)

Sounds good to me

cs3/storage/provider/v1beta1/resources.proto
ishank011: Your suggestion sounds good to me. Maybe we can wait for @labkode to take a look as well

butonic:
@ishank011 all updated

ishank011 (approved):

PR #131, opened 2021-06-11
Extend app registry with methods to add providers and mimetype filters

ishank011:
This PR adds methods to register new app providers on the fly and to set and retrieve default app providers for mime types. It has a change that might break the clients using the current version. We remove the mime type filters from `ListAppProvidersRequest` as that is redundant and performs the same functionality as `GetAppProvidersRequest`.

### DIFF
--- a/cs3/app/provider/v1beta1/provider_api.proto
+++ b/cs3/app/provider/v1beta1/provider_api.proto
@@ -89,12 +89,6 @@ message OpenInAppRequest {
   // The access token MUST be short-lived.
   // TODO(labkode): investigate token derivation techniques.
   string access_token = 4;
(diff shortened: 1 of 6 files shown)


glpatcern (commented):
Looks good to me: I see you simplified the filter for `ListAppProvidersRequest`, I had created it with the `oneof` form following the pattern of other similar requests, but if you think it's better like this I'm OK.

labkode:
@ishank011 fix the proto lock conflict. Maybe you'll need to create a new proto lock file. Please document breaking changes.

ishank011:
@labkode done

labkode (approved):

PR #134, opened 2021-06-21
New method PullTransfer.

redblom:
When a transfer type OCM share is accepted by the user at destination the transfer becomes a pull action. At that point all the info to start the actual transfer is available (at destination) and the resource is pulled from the source to destination.
`rpc PullTransfer(PullTransferRequest) returns (PullTransferResponse)`

See https://github.com/cs3org/reva/pull/1759#issuecomment-858516581

`PullTransferRequest` fields:
`targetURI` - holds all info regarding the resource at source and the access token
(see https://github.com/cs3org/reva/pull/1759#issuecomment-863094712)

datatx-ocmshare-pull-transfer-model

### DIFF
--- a/cs3/gateway/v1beta1/gateway_api.proto
+++ b/cs3/gateway/v1beta1/gateway_api.proto
@@ -365,12 +365,16 @@ service GatewayAPI {
   // ************************** FILE TRANSFER ************************/
   // *****************************************************************/
 
-  // Returns a response containing a TxInfo (transfer info) object.
-  rpc CreateTransfer(cs3.tx.v1beta1.CreateTransferRequest) returns (cs3.tx.v1beta1.CreateTransferResponse);
+  //  Requests the destination to pull a resource from source.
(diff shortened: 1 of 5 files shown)


redblom:
@Ishank, @labkode : About `ListTransfersRequest`
I gave it a filter field but I'm not sure how it should look like. I gave it a `status` field which would seem handy to me. But I think it would be practical eg. to be able to filter on shareID. 
Would this be correct?
```
message ListTransfersRequest {
  // OPTIONAL.
  // Opaque information.
  cs3.types.v1beta1.Opaque opaque = 1;
  // REQUIRED.
  // represents a filter to apply to the request.
  message Filter {
    // REQUIRED.
    oneof term {
      Status status = 1;
      // arbitrary filter term
    }
  }
  // OPTIONAL.
  // The list of filters to apply if any.
  repeated Filter filters = 2;
}
```

ishank011:
Instead of opaque, let's make the type concrete. For example, txID. Also please add the 'type' field to Filter, like we do for other requests.

redblom:
Superseded by https://github.com/cs3org/cs3apis/pull/143

PR #135, opened 2021-06-29
update the storageprovider recycle api

C0rby:
Added a reference for the target resource on which the action should be executed. This is necessary because we want to be able to perform actions for specific resources in the recycle file tree and not only on the root of a deleted tree.

I started in https://github.com/cs3org/reva/pull/1827 to extend the recycle/trash-bin APIs in a way that actions can be performed for specific resources in the trash-bin file tree.

It works so far but the API needs some changes.
- `ListRecycle` and `ListRecycleStream` had neither a key nor a reference
- `RestoreRecycleItem` had a key and a ref 

This PR makes the Recycle API related methods consistent by adding a ref where missing and removing the key since it can be included in the reference and so we don't need to separate fields.

### DIFF
--- a/cs3/storage/provider/v1beta1/provider_api.proto
+++ b/cs3/storage/provider/v1beta1/provider_api.proto
@@ -435,6 +435,9 @@ message ListRecycleRequest {
   // The end time range to query for recycle items.
   // The value is Unix Epoch timestamp in seconds.
   cs3.types.v1beta1.Timestamp to_ts = 3;
+  // OPTIONAL.
+  // The reference to which the action should be performed.
+  Reference ref = 4;
 }
 
 message ListRecycleResponse {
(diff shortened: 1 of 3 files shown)


cs3/storage/provider/v1beta1/provider_api.proto
```diff
@@ -545,9 +551,6 @@ message RestoreRecycleItemRequest {
   // REQUIRED.
   // The reference to which the action should be performed.
   Reference ref = 2;
-  // REQUIRED.
```
labkode, line 548: keep key

labkode:
Hi @C0rby,
looks good, only two things: 

a) keep the `key` field
b) do not alter the order, add reference at the end so change is backwards compatible.

For a).
EOS exposes a key and a reference for the root recycle entry ONLY.
Is not possible to restore subtrees in a consistent way.

A restore `key` is just a pointer to a collection of `references`.
For the ocfs driver, you can use the `key` to point to the root recycle entry and relative references to navigate the subtree.

Using the `reference` as a key is a hack and will add more complexity to the code when restoring a full root recycle entry vs some children of it.

C0rby:
Thanks for the feedback. I understand and will update my changes. :+1:

labkode (approved):

PR #136, opened 2021-07-02
add DenyGrant method

labkode:


refs (approved):

ishank011:
@labkode when I discussed this with Giuseppe, I think we talked about passing the `ResourcePermissions` object with all the fields set to false in case we want to deny access to a particular grantee. Are we not using that approach?

labkode:
@ishank011 is not enough. The empty resource permissions means "I don't have access" but we don't a way to express in the UI "I have powers to deny". So I've added this new method to the resource permissions to indicate that a person with this bit set can Deny access to others and hence the new method. When a user has the DenyGrant permission it will be translated to OCIS into a new WebDAV permissions to show the Deny role to the user.

ishank011:
Ah okay, I misunderstood that. Sounds good. I think the list of permissions is sorted alphabetically so let’s maintain that and I guess we can merge it then

ishank011:
Also, the rpc would need to be added to the gateway API as well

labkode:
@ishank011 true, I missed that, let me add it.
This is big refactor: https://github.com/cs3org/reva/pull/1856

labkode:
@ishank011 just realized is not needed in the gateway this is internal business.
We rely on CreateShare with empty permissions to call internally the DenyGrant.
Same reason we don't expose the AddGrants, ...
This is good to go

ishank011 (approved):

PR #138, opened 2021-07-29
Add the ListSupportedMimeTypes call

ishank011:
cc @wkloucek


cs3/app/registry/v1beta1/registry_api.proto
```diff
@@ -120,6 +122,24 @@ message ListAppProvidersResponse {
   repeated ProviderInfo providers = 3;
 }
 
+message ListSupportedMimeTypesRequest {
+  // OPTIONAL.
+  // Opaque information.
+  cs3.types.v1beta1.Opaque opaque = 1;
+}
+
+message ListSupportedMimeTypesResponse {
```
wkloucek, line 131: at least for the Web frontend we would need the app_name in here because this does not equal the provider_name!?

ishank011: Yep, I'll correct the naming. It actually stores the name of the app only

labkode (approved):

PR #139, opened 2021-08-03
Use a URL object in OpenInAppResponse

ishank011:
cc @wkloucek

### DIFF
--- a/cs3/app/provider/v1beta1/resources.proto
+++ b/cs3/app/provider/v1beta1/resources.proto
@@ -1,0 +1,46 @@
+// Copyright 2018-2019 CERN
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// In applying this license, CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+syntax = "proto3";
+
+package cs3.app.provider.v1beta1;
+
+option csharp_namespace = "Cs3.App.Provider.V1Beta1";
+option go_package = "providerv1beta1";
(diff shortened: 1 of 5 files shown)


cs3/app/provider/v1beta1/resources.proto
```diff
@@ -0,0 +1,46 @@
...
+option php_namespace = "Cs3\\App\\Provider\\V1Beta1";
+
+// Represents the information for the app URL to be called by the clients.
+message OpenInAppURL {
+  // REQUIRED.
+  // The URL that clients will perform requests to.
+  string app_url = 1;
+  // REQUIRED.
+  // The method for the request to be made.
+  string method = 2;
```
glpatcern, line 38: Looks all good to me. Do you think we need to restrict this to either `GET` or `POST`?

ishank011: Yep, makes sense. I'll add a comment

glpatcern (approved):

cs3/app/provider/v1beta1/resources.proto
```diff
@@ -0,0 +1,47 @@
...
+  // The method for the request to be made.
+  // Only GET and POST are supported.
+  string method = 2;
+  // OPTIONAL.
+  // The form parameters which have to be passed along with the request.
+  // These are sent only if the method is 'POST'.
+  map<string, string> form_parameters = 3;
+  // OPTIONAL.
+  // The headers to be added to the request.
+  map<string, string> headers = 4;
```
wkloucek, line 46: I read up on forms a little bit. The `GET` method on a form submit will add the form fields to the query parameters in the URL whereas the `POST` method will put them in the body (https://developer.mozilla.org/en-US/docs/Learn/Forms/Sending_and_retrieving_form_data).

But in neither case there is an explicit option to set Headers. How would it be used then?

ishank011: It is possible to pass headers to iframes, albeit not in a straightforward manner https://stackoverflow.com/a/60571237. The apps which we're considering don't need any to be set but it would be a good functionality to have IMO.

wkloucek (approved):

ishank011:
I've added a couple of fields to `ProviderInfo` to store the icon URI and a `desktop_only` parameter based on [this requirement](https://github.com/cs3org/reva/issues/1923#issuecomment-892703187). Now we return the complete provider object instead of just the name so the clients can use the icon field and we can filter these out in the HTTP layer based on the request's user-agent.

glpatcern:
As discussed, I'm merging this given that we tested the whole chain and we are ready to merge https://github.com/cs3org/reva/pull/1968 as well.

PR #141, opened 2021-08-13
add method to remove references to the storageprovider api

C0rby:
Add a new method to the storage provider to be able to remove references.
This is helpful when a user declines a share, since we only want to remove the reference to the shared file but not the file itself.

Once merged https://github.com/cs3org/reva/pull/1991 would make use of this change.

### DIFF
--- a/cs3/storage/provider/v1beta1/provider_api.proto
+++ b/cs3/storage/provider/v1beta1/provider_api.proto
@@ -138,6 +138,8 @@ service ProviderAPI {
   // Creates a reference to another resource in the same cluster or another domain (OCM shares).
   // The references resource can be accessed by the protocol specificied in the request message.
   rpc CreateReference(CreateReferenceRequest) returns (CreateReferenceResponse);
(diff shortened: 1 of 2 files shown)


labkode:
Hey @C0rby. A reference is just like a symbolic link and can be removed with the Remove operation.
When a share gets accepted, a new reference is created. When a share is declined the reference is removed with Remove operation. That is my understanding.

refs:
> Hey @C0rby. A reference is just like a symbolic link and can be removed with the Remove operation.
> When a share gets accepted, a new reference is created. When a share is declined the reference is removed with Remove operation. That is my understanding.

Hm, I was [under the same impression](https://cs3org.github.io/cs3apis/#cs3.storage.provider.v1beta1.DeleteRequest)

C0rby:
Ah, I see. Ok then I'll try to make https://github.com/cs3org/reva/pull/1991 work with the `Delete` method. :+1:

C0rby:
Ok, using the `Delete` method works. :+1:

PR #145, opened 2021-09-10
Extensions for the AppProviders to improve the UI

glpatcern:
This PR introduces further extensions for the open-in-app workflow:
* `AppProvider`s expose a `capability` to identify fully fledged editors vs viewers
* Mime types are defined including a friendly name and a corresponding file extension

This allows the UI to e.g. populate a **New** menu with all types for which an `editor` AppProvider is defined.

The extensions constitute breaking changes for the protobuf bindings, and can be merged once cs3org/reva#2067 is ready to be merged.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,8 @@ pull:
 
 build: pull
 	docker run -v ${pwd}:/root/cs3apis cs3org/cs3apis cs3apis-build -build-proto
+	# restore ownership of the `docs` folder as docker runs as root
+	chown -R `ls -ld . | awk '{print $$3 ":" $$4}'` docs
 python: pull
 	docker run -v ${pwd}:/root/cs3apis cs3org/cs3apis cs3apis-build -build-python
 go: pull
--- a/cs3/app/provider/v1beta1/provider_api.proto
+++ b/cs3/app/provider/v1beta1/provider_api.proto
@@ -53,7 +53,7 @@ import "cs3/types/v1beta1/types.proto";
 // Any method MAY return UNKNOWN.
 // Any method MAY return UNAUTHENTICATED.
 service ProviderAPI {
-  // Returns the App provider URL
+  // Returns the App URL and all necessary info to open a resource in an online editor.
   // MUST return CODE_NOT_FOUND if the resource does not exist.
   rpc OpenInApp(OpenInAppRequest) returns (OpenInAppResponse);
 }
--- a/cs3/app/registry/v1beta1/registry_api.proto
+++ b/cs3/app/registry/v1beta1/registry_api.proto
@@ -57,7 +57,7 @@ service RegistryAPI {
   rpc AddAppProvider(AddAppProviderRequest) returns (AddAppProviderResponse);
   // Returns a list of the available app providers known by this registry.
   rpc ListAppProviders(ListAppProvidersRequest) returns (ListAppProvidersResponse);
-  // Returns a list of the supported mime types along with the apps which they can opened with.
+  // Returns a list of the supported mime types along with the apps which they can be opened with.
   rpc ListSupportedMimeTypes(ListSupportedMimeTypesRequest) returns (ListSupportedMimeTypesResponse);
   // Returns the default app provider which serves a specified mime type.
   rpc GetDefaultAppProviderForMimeType(GetDefaultAppProviderForMimeTypeRequest) returns (GetDefaultAppProviderForMimeTypeResponse);
@@ -136,8 +136,8 @@ message ListSupportedMimeTypesResponse {
   // Opaque information.
   cs3.types.v1beta1.Opaque opaque = 2;
   // REQUIRED.
-  // The list of supported mime types with the apps which they can be opened with.
-  map<string, AppProviderList> mime_types = 3;
+  // The list of supported mime types and their properties.
+  repeated MimeTypeInfo mime_types = 3;
 }
 
 message GetDefaultAppProviderForMimeTypeRequest {
--- a/cs3/app/registry/v1beta1/resources.proto
+++ b/cs3/app/registry/v1beta1/resources.proto
@@ -42,20 +42,30 @@ message ProviderInfo {
   // The address where the app provider can be reached.
   // For example, tcp://localhost:1099.
   string address = 3;
+  // REQUIRED.
+  // The capability of the underlying app.
+  enum Capability {
+    CAPABILITY_INVALID = 0;
+    // The app is a simple viewer.
+    CAPABILITY_VIEWER = 1;
+    // The app is a full editor.
+    CAPABILITY_EDITOR = 2;
+  }
+  Capability capability = 4;
   // OPTIONAL.
-  // A human-readable name of the app provider.
-  string name = 4;
+  // A human-readable name of the underlying app.
+  string name = 5;
   // OPTIONAL.
   // Information to describe the functionalities
   // offered by the app provider. Meant to be read
   // by humans.
-  string description = 5;
+  string description = 6;
   // OPTIONAL.
   // A URI to a static asset which represents the app icon.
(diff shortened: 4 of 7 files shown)


labkode (approved):
Looks good. In general is better to not break backwards compatibility by maintaining the order of the fields in the protobuf definitions. However as this PR introduces a new REQUIRED fields it does not change anything.

glpatcern:
Yeah I thought about that and also about leaving them OPTIONAL but in the end we need them...

cs3/app/registry/v1beta1/resources.proto
```diff
@@ -42,23 +42,54 @@ message ProviderInfo {
...
   // For example, tcp://localhost:1099.
   string address = 3;
+  // REQUIRED.
+  // The capability of the underlying app.
+  enum Capability {
+    CAPABILITY_INVALID = 0;
+    // The app is a simple viewer.
+    CAPABILITY_VIEWER = 1;
+    // The app is a full editor.
+    CAPABILITY_EDITOR = 2;
```
wkloucek, line 52: How exactely would the UI make use of the capabilities? 
If it has `CAPABILITY_VIEWER` obviously shouldn't offer to edit a file... But if something is a `CAPABILITY_EDITOR` a user might also want to know if it's possible to open a file in reading mode.
Shouldn't the Capabilities reflect all possible OpenInAppRequest.ViewMode(s) https://cs3org.github.io/cs3apis/#cs3.gateway.v1beta1.OpenInAppRequest.ViewMode?

glpatcern: This was discussed in a now merged PR. There are two motivations for this:
* In the context menu for files, one could show `Preview in <app>` vs both `Open in <app>` and `Preview`. Of course if the user has no write access, only `Preview` would be shown.
* Populate the "New" button menu with mime types for which an AppProvider is registered with `EDITOR` capability. As a (counter-)example, assuming that e.g. Collabora is able to open jpeg files in view mode, do not show it in the New list.

Now the first is not super urgent in the current state of integrations, whereas the second is quite critical. The other part of this PR - providing a list of "well-known" mime types with their description and corresponding file extensions - specifically targets that.

Finally, I did think about some "beauty of symmetry" to make `capabilities` reflect the `viewMode` values, but I don't see the added value - for the UI you just need to know if an app is capable of editing or not, a viewer with or without a download button (and with or without the ability to hide it should the permission be so restrictive) is still a viewer and would be fired from a `Preview` menu item.

glpatcern: Another upfront consideration is the fact that we do have a number of "UI-only" apps (e.g. drawio, the pdfviewer, etc.) which are not captured by the AppProviders scheme. So anyway the UI must take those into account differently.

glpatcern:
Self-reviewing: this model is still missing the ability to register the mime type attributes (name, ext, icon...) from "outside" (e.g. the configuration). A `SetMimeType` RPC method may be needed for this.

Also, I don't understand why such a large diff in `proto.lock`. I dropped and recreated it as this is a breaking change, does that mean that we were carrying over lots of "garbage"?

Makefile
```diff
@@ -8,6 +8,7 @@ pull:
 
 build: pull
 	docker run -v ${pwd}:/root/cs3apis cs3org/cs3apis cs3apis-build -build-proto
+	chown -R `ls -ld . | awk '{print $$3 ":" $$4}'` docs
```
labkode, line 11: Explanation please :)

cs3/app/provider/v1beta1/provider_api.proto
```diff
@@ -53,11 +53,38 @@ import "cs3/types/v1beta1/types.proto";
...
+  // OPTIONAL.
+  // Opaque information.
+  cs3.types.v1beta1.Opaque opaque = 1;
+  // REQUIRED.
+  // The target container where the file has to be created.
+  cs3.storage.provider.v1beta1.Reference ref = 2;
+  // REQUIRED.
+  // The name of the file to be created.
+  string filename = 3;
+}
```
labkode, line 74: You can add an optional field named "template". Looking forward in case we want to provide custom templates (like CERN branded Collabora files).

Makefile
glpatcern: :-D sure - I'll add a comment in the file

cs3/app/provider/v1beta1/provider_api.proto
```diff
@@ -53,11 +53,38 @@ import "cs3/types/v1beta1/types.proto";
 // Any method MAY return UNKNOWN.
 // Any method MAY return UNAUTHENTICATED.
 service ProviderAPI {
-  // Returns the App provider URL
+  // Creates a new file to be opened by an app. The new file MAY be either
+  // empty or generated out of a template if the AppProvider has such capability.
+  rpc CreateFileForApp(CreateFileForAppRequest) returns (CreateFileForAppResponse);
```
labkode, line 58: From the UI perspective. As a I user I should be able to create for example: "New MS Word document" or "New OnlyOffice document". I think we may need to add the `app` name in the payload.

cs3/app/provider/v1beta1/provider_api.proto
glpatcern: Makes sense, as eventually the user chooses between e.g. `New Word Document` or `New Word Document from corporate template`...

cs3/app/provider/v1beta1/provider_api.proto
glpatcern: Just discussed with Lisa and Diogo: the creation of the file (either touch or copy from template) takes place before firing the app, and it does not involve the app at all - at least, so far we don't know any app (definitely not the WOPI ones) that expose an endpoint to _create_ a new file. The new file/template operation is therefore only on the storage and app independent.

glpatcern:
Will merge after https://github.com/cs3org/cs3apis/pull/145 works

cs3/app/registry/v1beta1/resources.proto
wkloucek: > Another upfront consideration is the fact that we do have a number of "UI-only" apps (e.g. drawio, the pdfviewer, etc.) which are not captured by the AppProviders scheme. So anyway the UI must take those into account differently.

That's true but these apps are also ownCloud Web specific and can't be used in any other web frontend. The app driver returns generic application which can also be used in a different web frontend. Therefore it is ok to just list what the backend / reva provides.

cs3/app/registry/v1beta1/registry_api.proto
```diff
@@ -136,8 +136,8 @@ message ListSupportedMimeTypesResponse {
   // Opaque information.
   cs3.types.v1beta1.Opaque opaque = 2;
   // REQUIRED.
-  // The list of supported mime types with the apps which they can be opened with.
-  map<string, AppProviderList> mime_types = 3;
+  // The list of supported mime types and their properties.
+  MimeTypeList mime_types = 3;
```
ishank011, line 140: Instead of `MimeTypeList`, we can just use `repeated MimeTypeInfo` here. What do you think?

ishank011:
Also, you'll need to add the proto lock file for the build to pass

cs3/app/registry/v1beta1/registry_api.proto
glpatcern: It's equivalent indeed, but isn't it convenient to have a type for the list in Reva? or in that case we just have a `map`?

ishank011: Yep, we had to use this with a map because `map<string, repeated message>` isn't supported. It adds one more layer of abstraction which doesn't have much use. The access now would look like `response.MimeTypes.MimeTypes`, when we can have `response.MimeTypes`.

PR #147, opened 2021-09-23
Add reference to GetQuota rpc

micbar:
# Description

- To query also the quota of a storage space, we need an optional parameter in the GetQuota rpc


refs (approved):

butonic:
@micbar can you rebase?

micbar (commented):
  cs3.storage.provider.v1beta1.Reference ref

Use a reference to pass on to the storage provider

micbar:
rebased and using the reference now.

ishank011 (approved):

PR #149, opened 2021-10-06
Add a create flag to MimeTypeInfo

glpatcern:
This flag is to be consumed by the web UI: when true, the given mime type is considered eligible for file creation and shall be shown in the New button. Default is false.


ishank011:
Maybe rename to `allow_creation`?

PR #154, opened 2021-10-28
add TouchFile

wkloucek:
## Description
~`CreateFile`~ `TouchFile` can be used to create an empty file. This is the file counterpart of `CreateContainer`. It allows one to easily create a file in order to reference it with a storage and opaque id. (Referencing it with a path or a relative path is already possible without the existing file).

## Alternative considerations
One could use the `InitiateFileUpload` in order to upload a zero byte file. it will give you an array of upload methods (eg. TUS and simple) and you can then invoke the upload of the zero byte file.

This means you need an http request ( in case of the simple protocol ) in addition to the InitiateFileUpload grpc request.

To spare that http request, some file system implementations, eg the decomposedfs implement a shortcut (https://github.com/cs3org/reva/blob/4879bf6be97aecc9e30ad999355342b6ca6188f1/pkg/storage/utils/decomposedfs/upload.go#L300-L308) for zero byte files. The cs3 api client has no way to check that the upload already finishend, because there are no return codes. Therefore bugs like https://github.com/owncloud/ocis/issues/2609 arise. 
In my opinion, this shortcut for zero byte files is also wrong, because the InitiateFileUpload is only an intend to upload a file. The actual file should created when the file is finally uploaded by using the offered upload protocols (eg. simple or TUS). This also applies to zero byte files.

### DIFF
--- a/cs3/gateway/v1beta1/gateway_api.proto
+++ b/cs3/gateway/v1beta1/gateway_api.proto
@@ -99,7 +99,7 @@ service GatewayAPI {
   // Creates a new resource of type file.
   // MUST return CODE_PRECONDITION_FAILED if the file
   // cannot be created at the specified reference.
-  rpc CreateFile(cs3.storage.provider.v1beta1.CreateFileRequest) returns (cs3.storage.provider.v1beta1.CreateFileResponse);
+  rpc TouchFile(cs3.storage.provider.v1beta1.TouchFileRequest) returns (cs3.storage.provider.v1beta1.TouchFileResponse);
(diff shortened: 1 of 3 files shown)


labkode:
@wkloucek looks good, however can you rename the method to `TouchFile`? CreateFile may imply that file contains contents and is not the case.

labkode (changes requested):
Chang of method name to TouchFile

ishank011 (approved):

refs (approved):

PR #157, opened 2021-11-03
add default_application to MimeTypeInfo

wkloucek:
needed by https://github.com/cs3org/reva/pull/2230

### DIFF
--- a/cs3/app/registry/v1beta1/resources.proto
+++ b/cs3/app/registry/v1beta1/resources.proto
@@ -94,7 +94,7 @@ message MimeTypeInfo {
   // OPTIONAL.
   // Whether the mime type is eligible for file creation in the web UI.
   // Defaults to false, i.e. files with this mime type can be opened
-  // but not directly allow_creationd from the web UI.
+  // but not directly created from the web UI.
   bool allow_creation = 8;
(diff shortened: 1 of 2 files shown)


cs3/app/registry/v1beta1/resources.proto
```diff
@@ -96,4 +96,7 @@ message MimeTypeInfo {
   // Defaults to false, i.e. files with this mime type can be opened
   // but not directly allow_creationd from the web UI.
```
glpatcern, line 97:
```suggestion
  // but not directly created from the web UI.
```

glpatcern (approved):
Looks good to me, can you please also apply the (unrelated) suggested change?

glpatcern:
Thanks Willy, I don't get why the build is failing though: could you perhaps try and `make` so all artifacts are regenerated?

wkloucek:
> Thanks Willy, I don't get why the build is failing though: could you perhaps try and `make` so all artifacts are regenerated?

I already posted in the CS3apis Gitter channel about this. I think the changes fromhttps://github.com/cs3org/cs3apis-build/pull/18 broke it.

wkloucek:
@glpatcern CI is green now

PR #159, opened 2021-11-17
Clarify UserId idp should be an origin

michielbdejong:
As discussed in https://github.com/cs3org/reva/pull/2285

### DIFF
--- a/cs3/identity/user/v1beta1/resources.proto
+++ b/cs3/identity/user/v1beta1/resources.proto
@@ -33,7 +33,7 @@ import "cs3/types/v1beta1/types.proto";
 // A UserId represents a user.
 message UserId {
   // REQUIRED.
-  // The identity provider for the user.
+  // The origin (url without path) of the identity provider for the user, e.g. "https://example.com" or "http://localhost:8080"
   string idp = 1;
   // REQUIRED.
   // the unique identifier for the user in the scope of


refs (approved):

cs3/identity/user/v1beta1/resources.proto
```diff
@@ -33,7 +33,7 @@ import "cs3/types/v1beta1/types.proto";
 // A UserId represents a user.
 message UserId {
   // REQUIRED.
-  // The identity provider for the user.
+  // The origin (url without path) of the identity provider for the user, e.g. "https://example.com" or "http://localhost:8080"
```
wkloucek, line 36: how about IDPs supporting multiple realms like eg. Keycloak?  (see also https://www.keycloak.org/docs/latest/server_admin/#core-concepts-and-terms)

Example:
`https://keycloak.owncloud.test/auth/realms/oCIS-realm-1/protocol/openid-connect/...`
`https://keycloak.owncloud.test/auth/realms/oCIS-realm-2/protocol/openid-connect/...`

The realms `oCIS-realm-1` and `oCIS-realm-2` behave like totally separate IDPs.
Maybe both of them have a user "albert", should they both be treated as the same user in REVA?

I don't think so. Therefore we would need to keep the full URL with path.

micbar:
closing, see @wkloucek

PR #161, opened 2021-12-08
first prototype for the permissions service api

C0rby:
This is the proposed CS3 API extension to introduce a permissions service.
Currently this API is read-only but eventually should also allow to set or update permissions / assignments.

### DIFF
--- a/cs3/permissions/v1beta1/permissions_api.proto
+++ b/cs3/permissions/v1beta1/permissions_api.proto
@@ -1,0 +1,59 @@
+// Copyright 2018-2021 CERN
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


refs:
@ishank011 could you review this one please?

cs3/permissions/v1beta1/permissions_api.proto
```diff
@@ -0,0 +1,65 @@
...
+
+import "cs3/permissions/v1beta1/resources.proto";
+import "cs3/rpc/v1beta1/status.proto";
+import "cs3/storage/provider/v1beta1/resources.proto";
+
+// PermissionsAPI defines a service for permissions.
+service PermissionsAPI {
+  // CheckPermission defines a method to check permission/role.
+  rpc CheckPermission(CheckPermissionRequest) returns (CheckPermissionResponse);
+  // Could be added later on:
```
refs, line 39: I think we can get rid of this comment here

refs:
@C0rby rebase? 👼

refs:
Reva implementation lives in https://github.com/cs3org/reva/pull/2341

C0rby:
Ah, I messed something up. There are changes in this PR which are not from me. I'll fix that and then it should be good to go.

C0rby:
I don't really know why this is here: https://github.com/cs3org/cs3apis/pull/161/files#diff-b04b38d4e36f7a7171aeb211bf933eaf36d41d9866ebbc3639f673f84dc350aeL5710-R5906.

Rebasing didn't remove it. I don't find any place where I am touching this in the proto files..

PR #163, opened 2022-02-09
Add an optional lock_id argument to storage methods

glpatcern:
Follow up of cs3org/cs3apis#162: the storage methods that modify a resource now have an optional `lock_id` argument, such that storage provider implementations can enforce the lock validation.

In addition, also the `InitiateDownloadRequest` has this extra argument, in order to honor exclusive locks. The assumption then is that stat and metadata read operations are always allowed (when access rights are met), such that e.g. the web UI can show users that "File important_stuff.txt is locked in exclusive mode by user bob".

### DIFF
--- a/cs3/storage/provider/v1beta1/provider_api.proto
+++ b/cs3/storage/provider/v1beta1/provider_api.proto
@@ -214,6 +214,10 @@ message AddGrantRequest {
   // REQUIRED.
   // The grant to be added.
   Grant grant = 3;
+  // OPTIONAL.
+  // A lock_id: should the reference be locked, the stored
+  // lock_id MUST be equal to the given value.
+  string lock_id = 4;
 }
 
 message AddGrantResponse {
@@ -235,6 +239,10 @@ message DenyGrantRequest {
   // REQUIRED.
   // The grantee to remove permission.
   Grantee grantee = 3;
+  // OPTIONAL.
+  // A lock_id: should the reference be locked, the stored
+  // lock_id MUST be equal to the given value.
+  string lock_id = 4;
 }
 
 message DenyGrantResponse {
@@ -289,6 +297,10 @@ message DeleteRequest {
   // REQUIRED.
   // The reference to which the action should be performed.
   Reference ref = 2;
+  // OPTIONAL.
+  // A lock_id: should the reference be locked, the stored
+  // lock_id MUST be equal to the given value.
+  string lock_id = 4;
 }
 
 message DeleteResponse {
@@ -366,6 +378,10 @@ message InitiateFileUploadRequest {
     // return CODE_FAILED_PRECONDITION.
     string if_match = 4;
   }
+  // OPTIONAL.
+  // A lock_id: should the reference exist and be locked, the stored
+  // lock_id MUST be equal to the given value.
+  string lock_id = 5;
 }
 
 message InitiateFileUploadResponse {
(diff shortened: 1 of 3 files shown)


cs3/storage/provider/v1beta1/provider_api.proto
```diff
@@ -665,6 +689,10 @@ message RemoveGrantRequest {
   // REQUIRED.
   // The grant to remove.
   Grant grant = 3;
+  // OPTIONAL.
+  // A lock_id: should the reference be locked and the lock be exclusive,
+  // the stored lock_id MUST be equal to the given value.
+  string lock_id = 4;
```
butonic, line 695: I'm not entirely sure if move/copy need a source as well as a destination lock. I'd merge like this and add another lock_id  in a subsequent PR if necessary.

cs3/storage/provider/v1beta1/provider_api.proto
```diff
@@ -796,6 +832,10 @@ message UnsetArbitraryMetadataRequest {
   // REQUIRED.
   // The arbitrary metadata to delete.
   repeated string arbitrary_metadata_keys = 3;
+  // OPTIONAL.
+  // A lock_id: should the reference be locked, the stored
+  // lock_id MUST be equal to the given value.
+  string lock_id = 4;
 }
 
 message UnsetArbitraryMetadataResponse {
```
butonic, line 841: unset also needs a lock_id parameter

butonic (changes requested):
I think these messages also need a `lock_id`:
- [ ] CreateContainer
- [ ] TouchFile
- [ ] RestoreFileVersion
- [ ] RestoreRecycleItem

they all might end up writing into a locked resource

glpatcern:
We looked at those too:
- `CreateContainer` and `TouchFile` expect to create a resource and must fail if it exists
- Restore actions indeed may need the `lock_id` argument, going to add it

glpatcern:
I ended up adding the `lock_id` argument also to `OpenInApp`, though this is far-fetching: it would be needed when exclusive locks are implemented.

glpatcern:
@butonic not sure why I can't ask you to review again, anyway could you have another look? I'd go ahead and merge this as is, in case we can keep adding further `lock_id` arguments if we deem it needed.

ishank011:
Can we also rename `Unlock` to `ReleaseLock`?

ishank011 (commented):
For `RestoreFileVersion` and `RestoreRecycleItem`, we should specify that the lock is for the restore path

cs3/storage/provider/v1beta1/provider_api.proto
```diff
@@ -289,6 +297,10 @@ message DeleteRequest {
   // REQUIRED.
   // The reference to which the action should be performed.
   Reference ref = 2;
+  // OPTIONAL.
+  // A lock_id: should the reference be locked, the stored
+  // lock_id MUST be equal to the given value.
+  string lock_id = 4;
```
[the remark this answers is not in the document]
glpatcern: oops, fixed

glpatcern:
> Can we also rename `Unlock` to `ReleaseLock`?

As discussed, `Unlock` came from the WOPI terminology, so we leave it as is.

Otherwise, I expanded the spec for `RestoreRecycleItem`.

glpatcern:
> Should we also enforce it for `List`, since it is equivalent in case the exclusive lock is for a directory?

As a matter of fact locking a directory was not taken into account so far. I think we should put a comment that although `SetLock` supports any reference, the semantics of locked containers are not specified at this stage.

ishank011 (approved):

PR #165, opened 2022-02-27
add grant methods to the gateway

micbar:
# Description

The methods to work with Grants are currently only available directly on the storageprovider.

This PR adds them to the Gateway too.

## Background

We are using a stat cache in the gateway which also caches the grants. Currently it it completely unaware when something regarding the grants changes because these methods are currently called directly on the storage providers bypassing the gateway.

## Implementation

I aim to implement this on `master` and `edge` branch because the change is relatively contained.

### DIFF
--- a/cs3/gateway/v1beta1/gateway_api.proto
+++ b/cs3/gateway/v1beta1/gateway_api.proto
@@ -203,6 +203,24 @@ service GatewayAPI {
   rpc UpdateStorageSpace(cs3.storage.provider.v1beta1.UpdateStorageSpaceRequest) returns (cs3.storage.provider.v1beta1.UpdateStorageSpaceResponse);
   // Deletes a storage space.
   rpc DeleteStorageSpace(cs3.storage.provider.v1beta1.DeleteStorageSpaceRequest) returns (cs3.storage.provider.v1beta1.DeleteStorageSpaceResponse);
+  // Adds a new grant for the provided reference.
+  // MUST return CODE_NOT_FOUND if the reference does not exist
+  rpc AddGrant(cs3.storage.provider.v1beta1.AddGrantRequest) returns (cs3.storage.provider.v1beta1.AddGrantResponse);
+  // Denies access to the provided reference.
+  // MUST return CODE_NOT_FOUND if the reference does not exist
+  rpc DenyGrant(cs3.storage.provider.v1beta1.DenyGrantRequest) returns (cs3.storage.provider.v1beta1.DenyGrantResponse);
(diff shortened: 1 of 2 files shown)


micbar:
@kobergj @butonic @dragotin

kobergj (approved):

butonic:
The idea of Grants is that they are internal representations of permissions on storage providers. If and how they are created is an implementetion detail of the Share manager: it persist the share intent, independently of the storage providers / drivers. The gatewaycan be configured to forward sharing requests as `Grant`s to the storage providers to give them a chance to also represent the share in whatever way is suitable, eg. ACLs.

@labkode correct me if the above is a misrepresentation from my side.

In https://github.com/cs3org/reva/pull/2250 @C0rby had to bypass the gateway to add members to spaces. I remember having a discussion with him about how to proceed, but I don't remember the reasons. Maybe he can chime in.

I am saying that because I don't like to expose two sets of apis (Shares & Grants) that seemingly are intended for the same thing: managing access to resoucres. This would be confusing.

labkode:
> The idea of Grants is that they are internal representations of permissions on storage providers. If and how they are created is an implementetion detail of the Share manager: it persist the share intent, independently of the storage providers / drivers. The gatewaycan be configured to forward sharing requests as `Grant`s to the storage providers to give them a chance to also represent the share in whatever way is suitable, eg. ACLs.
> 
> @labkode correct me if the above is a misrepresentation from my side.

You're right! The Sharing API should be the one exposed to clients to define share behaviour. 
Exposing the Grants directly does not bring any benefit as this API is intended to be internal only.

micbar:
IMO that does't make any sense. Bypassing the gateway makes it even worse.

micbar:
We need a solution for this. And a pragmatic one. Because currently we expose the grants as Permissions to the clients.
Therefore we are caching them in the gateway. But the gateway is completely unaware of permission changes. That is because we are "hiding" these operations internally.

micbar:
Ok, i thought you were talking about the OCS Shares HTTP API.

I was not aware of the CS3 Shares Api.

This fits better.

butonic:
@micbar @C0rby I think it makes more sense to add a list of managers to the StorageSpace message: https://cs3org.github.io/cs3apis/#cs3.storage.provider.v1beta1.StorageSpace

That would allow using UpdateSpace to change the managers.

C0rby:
What about a list of Grants to represent space members. Then the roles of the members are not defined by the list and we could define more roles in the future based on the permissions.

PR #167, opened 2022-03-24
Added parent_id reference to ResourceInfo

glpatcern:
As discussed in https://github.com/cs3org/reva/issues/1804#issuecomment-1076471437 this PR introduces a `parent_id` property in `ResourceInfo`.

The property is added as `OPTIONAL`, both to not introduce a breaking change and to keep compatibility with the case of absolute paths, where it is actually not needed.

@butonic and @wkloucek, do you agree on the spec? including the fact that for single-file shares it `MUST` be set to `0`?

### DIFF
--- a/cs3/storage/provider/v1beta1/resources.proto
+++ b/cs3/storage/provider/v1beta1/resources.proto
@@ -77,25 +77,32 @@ message ResourceInfo {
   // Is the accounting recursive?, could it be set to 0 for directories if recursive not supported? use another field?
   // TODO(moscicki): This needs to be defined also for other types (such as a symlink to a directory or file)
   uint64 size = 10;
-  // REQUIRED:
+  // REQUIRED.
   // Identifier of the owner of the resource.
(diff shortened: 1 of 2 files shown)


glpatcern:
The last commit specifies to populate the `parent_id` in all cases of relative paths, including when the user has no access to the parent: after all, ids are opaque, and a subsequent `stat()` on that reference would give permission denied. So nothing is really disclosed, but at least a reference where you have `storageID` == the space ID, `opaqueID` == `parentID` and `path` == `./thefilename` is legitimate. Otherwise, a single-file share won't have a proper representation in that form, the only valid one being `storageID` the share ID and `opaqueID` == `fileID` with no path, thus preventing to refer to that file by filename.

Does that sound reasonable? cc @butonic, @wkloucek and @labkode

labkode (approved):

labkode:
Hi, I've merged it as-is because I haven't found any blocker and this is critical for spaces deployment with WOPI.
I suggest we start from here and iron out details as we face new challenges.

PR #168, opened 2022-04-06
Add quicklink field to link resource

kobergj:
`quicklink` is a simple bool value that indicates a unique link. The server will enforce that there is at max 1 quicklink per resource. This way the client can be sure there are not more than 1 quicklink and can adjust its UI.


micbar:
@labkode @ishank011 can you please review and merge?

We are just releasing ocis with a fork of the cs3apis

ishank011 (approved):

PR #169, opened 2022-04-26
storageprovider: add filter for path

kobergj:
Adds a filter for type `PATH` to the ListStorageSpacesRequest

PR #170, opened 2022-05-12
Add path filter to list storage spaces request

ishank011:


ishank011:
Ah a PR already exists https://github.com/cs3org/cs3apis/pull/169, but it didn't generate the docs, so we can merge this one

PR #174, opened 2022-06-21
add spaceID to resourceID

micbar:
# Description

To correctly address storage spaces in reva, we need another id property on the resourceID.

This change is not breaking and should be compatible for the `master` and the `edge` branches.

1) `master`branch.

We do not need to change anything

2) `edge` branch

Currently the `storageID` attribute is used for a concatenated `<storageproviderID>$<spaceID>`

With this change, we will be able to separate them and have a clean implementation back.

@labkode @glpatcern I will also update the cs3apis on the master branch to keep everything on the same version of the apis.

### DIFF
--- a/cs3/storage/provider/v1beta1/resources.proto
+++ b/cs3/storage/provider/v1beta1/resources.proto
@@ -283,6 +283,11 @@ message ResourceId {
   // uniquely identity the resource in the internal
   // implementation of the service.
   string opaque_id = 2;
+  // OPTIONAL.
+  // The internal id used by service implementor to
+  // uniquely identify the storage space.
+  // Used by the storageprovider to locate the correct storage space.
+  string space_id = 3;
 }
 
 // The representation of permissions attached to a resource.
--- a/docs/index.html
+++ b/docs/index.html
@@ -17725,6 +17725,16 @@ uniquely identity the resource in the internal
 implementation of the service. </p></td>
                 </tr>
               
+                <tr>
+                  <td>space_id</td>
+                  <td><a href="#string">string</a></td>
+                  <td></td>
+                  <td><p>OPTIONAL.
+The internal id used by service implementor to
+uniquely identify the storage space.
+Used by the storageprovider to locate the correct storage space. </p></td>
+                </tr>
+              
             </tbody>
           </table>
 


cs3/storage/provider/v1beta1/resources.proto
```diff
@@ -283,6 +283,11 @@ message ResourceId {
   // uniquely identity the resource in the internal
   // implementation of the service.
   string opaque_id = 2;
+  // REQUIRED.
```
kobergj, line 286: I would argue that adding a new value that is `REQUIRED` IS indeed a breaking change. Also it will need some integration in ocis. The reason is:

We used `StorageID` (REQUIRED FIELD) as the spaceID before. After we added the StorageProviderID (OPTIONAL FIELD) we used this field for both so we had `StorageProviderID$SpaceID` or only `SpaceID`

Now that would be completely different: `StorageID` would now be the StorageProviderID or empty. While the new SpaceID field would be REQUIRED.

kobergj: Additionally I doubt that it would be compatible with edge without changes. Did you test that?

micbar: It is only "REQUIRED" on the edge branch. The master branch will not need to change.

> Now that would be completely different: StorageID would now be the StorageProviderID or empty. While the new SpaceID field would be REQUIRED.

Yes, exactly. That means a clean implementation of spaces in reva edge and ocis.

kobergj (approved):

cs3/storage/provider/v1beta1/resources.proto
butonic: The `REQUIRED` comment is not enforced. that is why this change will work as is in both: `master` and `edge`. In `edge` we can then get rid of the storageid splitting hack and continue with a clean api.

glpatcern: I agree with @kobergj that this IS a breaking change. Rather, I propose to leave the `storage_id` as `REQUIRED` and add the new `space_id` as `OPTIONAL`.

IMHO adding this field is a welcome extension we (I, at least) advocated, to allow having a cleaner and more self-explanatory code in Reva edge, but at the same time I'd argue that the whole concept of spaces, at least at the API level, is optional, as we can have implementations (Reva master) that are doing without them. That's why the `space_id` should be `OPTIONAL`.

micbar: Ok, let us rephrase it that it is a **needed** change. whether it is breaking or not. I meant not breaking because `master` needs no code changes.

micbar: Honestly with these `REQUIRED` and `OPTIONAL` values they are a bit pointless.

on the `master` branch `space_id` is optional
on the `edge` branch `storage_id` is optional

micbar: spaces are the future so we should make the `storage_id` optional.

I can also live with it vice versa. These flags have no effect on the code.

labkode:
@micbar I agree with @kobergj and @glpatcern that this field MUST be an optional field for time being.
Apart from ownCloud, no one is using the SpacesAPI for their reference implementation.

kobergj:
@micbar @butonic @labkode @glpatcern How about we add a new OPTIONAL `StorageProviderID` and use the StorageID as SpaceID in ocis? That is basically the way ocis does it now too anyways.

micbar:
> @micbar I agree with @kobergj and @glpatcern that this field MUST be an optional field for time being.
Apart from ownCloud, no one is using the SpacesAPI for their reference implementation.

That surprises me.

I thought we have an agreement that spaces are the first class citizens in the API.

micbar:
@labkode @glpatcern I changed the REQUIRED flag to OPTIONAL. Not a blocker for me.

Please let us continue here.

glpatcern:
@micbar I think it's the best option at this point in time, leaving the door open to both approaches. Will merge.

PR #178, opened 2022-08-16
Add a if_unmodified_since condition to InitiateFileUploadRequest

aduffeck:
This change adds an optional `if_unmodified_since` field to InitiateFileUploadRequest which allows for conditional uploads based on the mtime.


butonic (approved):

glpatcern (approved):
Looks good, even more generic than the `if_not_exists` flag: I guess the intended scope for this new flag is to implement/enforce it on the sync clients, where eventually the `if_not_exists` flag should also be enforced.

PR #182, opened 2022-09-28
add UnlockAndRelock

micbar:
# Description

Implements https://github.com/cs3org/wopiserver/issues/86

### DIFF
--- a/cs3/storage/provider/v1beta1/provider_api.proto
+++ b/cs3/storage/provider/v1beta1/provider_api.proto
@@ -167,6 +167,14 @@ service ProviderAPI {
   // or if the caller does not hold the lock.
   // The caller MUST have write permissions on the resource.
   rpc RefreshLock(RefreshLockRequest) returns (RefreshLockResponse);
+  // Releases a lock on the storage resource and immediately relocks the storage resource.
+  // MUST return CODE_NOT_FOUND if the reference does not exist.
+  // MUST return CODE_FAILED_PRECONDITION if the reference is not locked,
+  // if the caller does not hold the existing lock or if the lock id does not match.
+  // The caller MUST have write permissions on the resource.
+  // In addition, the implementation MUST ensure atomicity when multiple users
+  // concurrently attempt to unlock and relock.
+  rpc UnlockAndRelock(UnlockAndRelockRequest) returns (UnlockAndRelockResponse);
   // Unlocks a storage resource.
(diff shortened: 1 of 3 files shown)


micbar:
@glpatcern

cs3/storage/provider/v1beta1/provider_api.proto
```diff
@@ -1022,6 +1030,33 @@ message RefreshLockResponse {
...
+message UnlockAndRelockResponse {
+  // REQUIRED.
+  // The response status.
+  cs3.rpc.v1beta1.Status status = 1;
+  // OPTIONAL.
+  // Opaque information.
+  cs3.types.v1beta1.Opaque opaque = 2;
+  // REQUIRED.
+  // The lock metadata
+  Lock lock = 3;
```
glpatcern, line 1057: Why would you need to return the new lock data structure on response? After all this operation is similar to RefreshLock, the server could just say "done" or "failed".

micbar: ok, you are correct. The new refresh metadata is already available for the caller. will change.

glpatcern:
@micbar I managed to check the current behavior of Reva `master` with the eosfs storage driver, and in fact the `RefreshLock` implementation does incorporate the `UnlockAndRelock` case. After all, the only behavioral difference between the two is enforcing or not the check on the `lock_id` payload. The decomposedfs implementation does enforce it:

https://github.com/cs3org/reva/blob/edge/pkg/storage/utils/decomposedfs/node/locks.go#L179-L182

And the eosfs implementation does not (similarly to the xroot "eos native" legacy implementation of the wopiserver). This is why for us the Microsoft WOPI validator test suite always passed.

With those premises, I'm really not convinced that it makes sense to add an extra call to the CS3APIs with such a minor behavioral difference, given that:
* the current specs do require to check the lock holder and the expiration time to protect the `RefreshLock` (as well as the `Unlock`) call
* the specs do not require to enforce that check on the opaque `lock_id` payload

Technically, dropping those 3 lines in locks.go and reverting cs3org/wopiserver#85 will make you pass the tests, with full atomicity preserved, instead of the overhead of adding a new API where the implementation would need to be a clone of the existing `RefreshLock` code - except for those 3 lines - for all storage providers.

I'd like to have more voices from @butonic and @labkode prior to go on or not.

micbar:
fine for me. I was also looking into the implementation.

PR #183, opened 2022-09-29
add optional oldLock to RefreshLockRequest

micbar:
# Description

Follow up of #182 

We decided together with @glpatcern that the refreshLock operation can be extended to be used for an atomic
`UnlockAndRelock` operation.

### DIFF
--- a/cs3/storage/provider/v1beta1/provider_api.proto
+++ b/cs3/storage/provider/v1beta1/provider_api.proto
@@ -1011,6 +1011,11 @@ message RefreshLockRequest {
   // REQUIRED.
   // The lock metadata.
   Lock lock = 3;
+  // OPTIONAL.
+  // The existing lock id.
+  // This can be used to refresh the existing lock with a new lock while
+  // checking if the lock holder sends the correct existing lock_id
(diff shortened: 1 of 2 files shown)


cs3/storage/provider/v1beta1/provider_api.proto
```diff
@@ -1011,6 +1011,11 @@ message RefreshLockRequest {
   // REQUIRED.
   // The lock metadata.
   Lock lock = 3;
+  // OPTIONAL.
+  // The existing lock metadata.
+  // This can be used to refresh the lock with a new lock while
+  // ensuring that the lock holder also holds the existing lock
+  Lock existing_lock = 4;
```
glpatcern, line 1018: Makes sense. What about passing `string existing_lock_id` as opposed to the whole `Lock` structure? The `user` or `appname` fields must match in all cases.

micbar: fine.

glpatcern (approved):

glpatcern:
Just wanted to wait for the CI, then I'll merge!

PR #190, opened 2022-12-08
Add Status to ResourceInfo

kobergj:
Adds a `Status` field to the resource info of a resource. This field can be used to signal a meta status of the resource (like "processing", "readytouse", "blocked", ...). For backwards compatibility an empty status should be considered as "ok"

### DIFF
--- a/cs3/storage/provider/v1beta1/resources.proto
+++ b/cs3/storage/provider/v1beta1/resources.proto
@@ -109,6 +109,9 @@ message ResourceInfo {
   // OPTIONAL.
   // StorageSpace where this resource is located.
   StorageSpace space = 19;
+  // OPTIONAL
+  // Status of the resource
+  Status string = 20;
 }
 
 // CanonicalMetadata contains extra metadata


micbar:
@labkode @vascoguita The Generator seems to be broken, can you investigate why?

micbar:
@kobergj Can you rebase?

flash7777:
Hi @kobergj — we just opened #272 which also touches `ResourceInfo` field 20.

We are adding an `immutable` boolean (field 20) as a persistent attribute for freezing resources (DMS/Aktenplan use case). We see this as fundamentally different from a transient processing status:

- **Status** (this PR): system/app-driven, transient — "processing", "blocked", etc.
- **Immutable** (#272): admin-driven, persistent — resource cannot be modified/deleted until explicitly unfrozen via `UnsetImmutable` RPC

Since both target field 20, we should coordinate. A few options:

1. One of us moves to field 21
2. We combine both into a single message with both concepts
3. Status becomes an enum (as suggested in the review) and we add `RESOURCE_STATUS_IMMUTABLE` — though we believe immutable deserves its own field since it has dedicated RPCs (`SetImmutable`/`UnsetImmutable`) and different semantics

Happy to discuss and find the best path forward.

PR #192, opened 2023-01-09
Share expiration

C0rby:
Added changes to the API to support expiration on user and group shares.

### DIFF
--- a/cs3/sharing/collaboration/v1beta1/collaboration_api.proto
+++ b/cs3/sharing/collaboration/v1beta1/collaboration_api.proto
@@ -111,7 +111,7 @@ message UpdateShareRequest {
   // Opaque information.
   cs3.types.v1beta1.Opaque opaque = 1;
   // REQUIRED.
-
+  // Deprecated use the share and update_mask attribute.
   ShareReference ref = 2;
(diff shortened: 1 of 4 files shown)


labkode (approved):

labkode:
@C0rby PR is good we just need to run the CI in the new infra, @micbar can you enable the cs3apis as well to run in ownCloud's drone?

micbar:
@labkode @C0rby Done. But there is an issue, the pipeline needs a gh token to publish the generated code.

@vascoguita How should we proceed?

vascoguita:
> @labkode @C0rby Done. But there is an issue, the pipeline needs a public key to publish the generated code.
> 
> @vascoguita How should we proceed?

I'm looking into it.
Ideally, there would be no SSH key at all, specially in a PR-triggered workflow.
Anyone could get the private key by making a PR with `echo $SSH_KEY` in the .drone.yml file.

I'll try to migrate this workflow to GH actions this afternoon. If there is urgency, I could temporarily @micbar the ssh key for the ownCloud drone instance.

micbar:
@vascoguita Thank you for the quick reaction. Afternoon sounds good 😄

Issue #189, opened 2022-11-25
Change `datatx.PullTransfer` to `datatx.CreateTransfer`


redblom:
By design a transfer tool is used for the data transfers. Depending on the tool or configuration transfers can be executed using different methods (streaming, third-party copy push/pull). Currently in our cs3 reference implementation reva we use rclone which supports both streaming and third-party push transfer (https://github.com/cs3org/reva/pull/3491) and we can configure this.

I propose to change `PullTransfer` method from the `datatx` module to `CreateTransfer`.

glpatcern:
Fully agree. If you create the PR here and in Reva, we can review and merge them at the same time to not break compatibility as a method rename is a breaking change.

redblom:
Do you want me to push an updated `proto.lock` file also? I noticed it's lagging behind.

glpatcern:
> Do you want me to push an updated `proto.lock` file also? I noticed it's lagging behind.

Why not, thanks.

PR #195, opened 2023-01-16
Name change in module datatx: PullTransfer -> CreateTransfer

redblom:
Deals with https://github.com/cs3org/cs3apis/issues/189
This is a breaking change.

### DIFF
--- a/cs3/gateway/v1beta1/gateway_api.proto
+++ b/cs3/gateway/v1beta1/gateway_api.proto
@@ -400,8 +400,8 @@ service GatewayAPI {
   // ************************** FILE TRANSFER ************************/
   // *****************************************************************/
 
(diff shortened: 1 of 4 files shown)


glpatcern (approved):
This is the famous rename that got the CI changes in the way - sorry for that. I think it's all fine but can you rebase on top of current `main`? and also, I guess we have to merge this and the corresponding Reva code change at the same time, right?

redblom:
> This is the famous rename that got the CI changes in the way - sorry for that. I think it's all fine but can you rebase on top of current `main`? 

Done

> and also, I guess we have to merge this and the corresponding Reva code change at the same time, right?

Yes that would be https://github.com/cs3org/reva/pull/3553

glpatcern (approved):

PR #203, opened 2023-02-21
add an optional quota to the CreateHomeRequest

C0rby:
We want to be able to limit the quota of newly created users.


micbar (approved):
LGTM

butonic (approved):

PR #206, opened 2023-03-21
[docs-only] added comments to detail the mapping of OCM properties

glpatcern:
Following the discussions in https://github.com/cs3org/OCM-API/issues/66, this is to prepare the CS3APIs so that we clarify how the OCM properties `protocol`, `providerId` are actually represented by the more appropriate naming `protocols`, `shareId`, already introduced in previous PRs.

Therefore this change is purely on comments. The implementation will follow in Reva.


gmgigi96:
LGTM! I would wait to merge this though, until the changes in OCM become more concrete.

glpatcern:
Agreed!

PR #210, opened 2023-05-16
the code locked has been added

2403905:
# Description

This adds: CODE_LOCKED to the rpc codes
The ability to lock a resource is specific to some WebDAV servers. 
The HTTP 423 Locked error response code indicates that either the resources tentatively targeted by is locked, meaning it can't be accessed. 
HTTP Mapping: 423 Locked


micbar (approved):
LGTM

PR #211, opened 2023-05-25
Added/updated methods to manipulate OCM shares and invitees

glpatcern:
Added missing OCM core APIs for updating and deleting an OCM share, and added method to delete (or unfriend) an accepted user. Also updated and reshuffled some comments to link the CS3 APIs to the corresponding OCM endpoints.

The OCM `AccessMethods` were reset (this is a potential breaking change) and a `destination` was added to the `transferAccessMethod`.


cs3/ocm/core/v1beta1/ocm_core_api.proto
```diff
@@ -50,11 +52,20 @@ import "cs3/types/v1beta1/types.proto";
 // Any method MAY return UNKNOWN.
 // Any method MAY return UNAUTHENTICATED.
 service OcmCoreAPI {
-  // Creates a new ocm share.
+  // Creates a new OCM share. The share SHALL be sent to the remote system
+  // using the OCM API at:
+  // https://cs3org.github.io/OCM-API/docs.html?branch=v1.1.0&repo=OCM-API&user=cs3org#/paths/~1shares/post
   rpc CreateOCMCoreShare(CreateOCMCoreShareRequest) returns (CreateOCMCoreShareResponse);
+  // Updates an OCM share. This action SHALL be notified to the remote system
```
gmgigi96, line 59: Actually if we reach this, the remote is already been notified, as this is called in the "notified" side, i.e. the recipient provider of the share

cs3/ocm/core/v1beta1/ocm_core_api.proto
```diff
@@ -109,3 +120,55 @@ message CreateOCMCoreShareResponse {
...
+  // REQUIRED.
+  // Unique ID to identify the share at the consumer side.
+  string ocm_share_id = 2;
+  // OPTIONAL.
+  // Description for the share.
+  string description = 3;
+  // OPTIONAL.
+  // Consumer specific identifier of the user or group the provider wants to share the resource with.
+  // TODO: this field needs to represent either a user or group in the future, not only a user.
+  cs3.identity.user.v1beta1.UserId share_with = 4;
```
gmgigi96, line 137: do we expect this to change? I expect that you cannot change the user of a share, like normal shares. instead you unshare and share with a new user. same will apply for groups in future

gmgigi96 (changes requested):

cs3/ocm/core/v1beta1/ocm_core_api.proto
glpatcern: Absolutely correct - when going through all fields from `Create` I overlooked that this is also to be dropped

cs3/ocm/core/v1beta1/ocm_core_api.proto
glpatcern: Correct, my last commit puts the comments in the right place

gmgigi96 (approved):

gmgigi96:
@labkode can you merge?

PR #212, opened 2023-07-24
Added target to OpenInAppURL object

glpatcern:
Also added `action` to AppProviderInfo.

The exercise here is to demonstrate a different kind of integration, where the app lives on an external cloud and uses its own login and storage. Therefore it can't be embedded in an iframe and the integration is more loose, with the user driving the `Export to` and `Import from` actions as opposed to a simple `Open with`.

### DIFF
--- a/cs3/app/provider/v1beta1/resources.proto
+++ b/cs3/app/provider/v1beta1/resources.proto
@@ -69,8 +69,8 @@ enum ViewMode {
 // Defines the valid targets for an app URL.
 enum Target {
   TARGET_INVALID = 0;
-  // The app URL is to be opened on an iframe
+  // The app URL is to be opened within an iframe
   TARGET_IFRAME = 1;
-  // The app URL is to be opened by redirecting to a new page
+  // The app URL is to be opened on a new blank page
   TARGET_BLANK = 2;
 }
--- a/cs3/app/provider/v1beta1/resources.proto
+++ b/cs3/app/provider/v1beta1/resources.proto
@@ -44,6 +44,9 @@ message OpenInAppURL {
   // OPTIONAL.
   // The headers to be added to the request.
   map<string, string> headers = 4;
+  // REQUIRED.
+  // Whether the target for the app URL is an iframe or a new page.
+  Target target = 5;
 }
 
 // Defines the view modes.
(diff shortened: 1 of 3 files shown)


cs3/app/provider/v1beta1/resources.proto
```diff
@@ -44,6 +44,9 @@ message OpenInAppURL {
   // OPTIONAL.
   // The headers to be added to the request.
   map<string, string> headers = 4;
+  // REQUIRED.
+  // Whether the target for the app URL is an iframe or a new page.
+  Target target = 5;
```
wkloucek, line 49: I'm not sure how `target` fits in here. Eg. doing a POST to a new tab is technically not possible.

ownCloud Web has a option `WEB_OPTION_OPEN_APPS_IN_TAB` -> `Configures whether apps and extensions should generally open in a new tab. It seems to fulfill the same purpose? In the future, this could also be a per user setting.

cs3/app/provider/v1beta1/resources.proto
```diff
@@ -62,3 +65,12 @@ enum ViewMode {
   // then this mode MUST fall back to READ_WRITE.
   VIEW_MODE_PREVIEW = 4;
 }
+
+// Defines the valid targets for an app URL.
+enum Target {
+  TARGET_INVALID = 0;
+  // The app URL is to be opened on an iframe
+  TARGET_IFRAME = 1;
+  // The app URL is to be opened by redirecting to a new page
```
wkloucek, line 74: `redirecting to a new page` sounds a little bit like a `HTTP 302`. 

When looking at existing descriptions of the "blank" target (eg. https://www.w3schools.com/tags/att_a_target.asp), `Opens the app URL in a new window or tab` could be a better fit.

glpatcern: yep, the comment can be improved, `blank` better represents the intended behavior

cs3/app/provider/v1beta1/resources.proto
glpatcern: Yes, with this not all combinations are legal, but the scope is for *certain* applications to go to a new tab/page as opposed to an iframe. We are trying to integrate Overleaf, but the integration would work quite differently than WOPI-based apps, therefore it makes sense (and it is necessary for CSP reasons) to have the user go directly to the app's URL, rather than embedding it in an iframe.

wkloucek: Ah, ok this is currently not configure for oC Web. Because even if one configures Web to open everything in a new tab, it will open the app inside a iframe via the GET / POSt request.

@micbar this is highly interesting for future integrations

glpatcern:
@labkode can you also have a look?

labkode (approved):

PR #214, opened 2023-09-27
align owncloud master against main

dragonchaser:


glpatcern:
@diocas how did we implement hiding of shares in CERNBox without this?

kobergj (approved):

micbar (approved):

glpatcern (changes requested):
Following direct chats with @diocas and @gmgigi96 : the current implementation uses the `state` of a `ReceivedShare`, such that `REJECTED` shares are hidden, and others are visible.
This field seems then redundant, unless we want to express (received) shares that are `APPROVED` but hidden, or `REJECTED` but visible.
Also, an outgoing `Share` does not have a state and does not need to be hidden?

cs3/sharing/collaboration/v1beta1/resources.proto
```diff
@@ -70,6 +70,9 @@ message Share {
   // Optional.
   // The expiration time of the share.
   cs3.types.v1beta1.Timestamp expiration = 9;
+  // Optional.
+  // Hide share, default false
+  bool hide = 10;
```
glpatcern, line 75:
```suggestion
  bool hidden = 10;
```

(Assuming we want to keep it, though it would rather belong to the `ReceivedShare` message.

micbar:
@glpatcern 

As I understand it, it is a CERN feature request to have two flags on each share

1) Accepted
2) Hidden

micbar:
We already implemented it on the backend and the web client for the CERN merge.

glpatcern:
@micbar not sure the two flags were understood to be independent, i.e. with the 4 combinations being meaningful. Anyway even assuming so, that still concerns the `ReceivedShare`, not the `Share`, right?

micbar:
User story https://github.com/owncloud/web/issues/9531

glpatcern:
Thanks, understood. I see the idea is also to observe adoption of those combinations and potentially simplify in the future.

With that, I think it's fine to add a `hidden` bool field to the `ReceivedShare`, next to (and indeed independent from) `state`. The `Share` structure/message should not be affected instead. Or do I miss something else?

cs3/sharing/collaboration/v1beta1/resources.proto
butonic: @dragonchaser I agree with @glpatcern ... I'd also use `hidden`. And the `hidden` status of share is used by recipients to hide shares they are not interested in. So the property should live in ReceivedShare. Otherwise the hidden state is shared by all recipients of the share. ... hm we can take a look at the implementation. AFAIR storing the property on the share made sense because the received share did not exist, yet. Or because listing shares should omit hidden shares?

dragonchaser: Yes this is on my list, will tackle this as soon as @aduffeck is done with the ocm story. I fear this is going to be a rebase hell for both of us otherwise :)

dragonchaser:
I will move it to `ReceivedShare` then....

glpatcern (approved):
That's fine, once you're ready with reva/edge this can be merged - and we'll adapt reva/master as well.

dragonchaser:
We have a PR for edge: https://github.com/cs3org/reva/pull/4256

dragonchaser:
@glpatcern if you want you can merge it :)

PR #215, opened 2023-10-24
Decouple spaces API from storage

gmgigi96:
The spaces API are currently coupled with the storage. This PR decouple them from the storage, allowing a more flexible implementation.

### DIFF
--- a/cs3/gateway/v1beta1/gateway_api.proto
+++ b/cs3/gateway/v1beta1/gateway_api.proto
@@ -46,6 +46,7 @@ import "cs3/sharing/collaboration/v1beta1/collaboration_api.proto";
 import "cs3/sharing/link/v1beta1/link_api.proto";
 import "cs3/sharing/ocm/v1beta1/ocm_api.proto";
 import "cs3/storage/provider/v1beta1/provider_api.proto";
+import "cs3/storage/provider/v1beta1/spaces_api.proto";
 import "cs3/storage/provider/v1beta1/resources.proto";
 import "cs3/tx/v1beta1/tx_api.proto";
 import "cs3/types/v1beta1/types.proto";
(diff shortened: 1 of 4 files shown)


labkode:
@gmgigi96 looks good to me to split the spaces and the storage api into two separate entities as the spaces sit on top storage and have functionalities on top of it. 

This is a breaking change because of a new API, however it should be straighforward at the code level to change as the messages are the same. 

@cs3org/owncloud-team please review.

micbar:
Hmm, from my pov I consider a storage provider which has no spaces support as an invalid implementation.

labkode:
@micbar reva@master does not have Spaces support and is not a valid implementation :)

gmgigi96:
> Hmm, from my pov I consider a storage provider which has no spaces support as an invalid implementation.

@micbar I think were you store spaces is just a matter of implementation detail. For use is a DB, for you the storage itself. That's why this need to be separated.

micbar:
> @micbar reva@master does not have Spaces support and is a valid implementation :)

Correct. But as I understood you some years ago, that should only be temporary. You had the plan to converge with reva edge. Did this plan change?

labkode:
@micbar plan is to converge on web only, and look at future options later.

micbar:
> @micbar I think where you store spaces is just a matter of implementation detail. For us is a DB, for you the storage itself. That's why this need to be separated.

I think you are missing the details. Every Stat call or every Call to the storage provider needs to implement the spaces semantics. Resources are referenced by <StorageID><SpaceID><OpaqueID>.

Reva master does only path based requests. On Reva Edge, Spaces are a basic architectural concept of total decentralization. Reva Master has the concept that there is a global path which is the same for everybody. These two concepts are different in its core.

Maybe I am the only one who is concerned about diverging too much.

@butonic @dragotin @kobergj @dragonchaser

PR #227, opened 2024-03-05
Further clarified scope of lock_id

glpatcern:
Follow up from #226: the `lock_id` payload of storage requests that do not alter the content of a file MAY or MAY NOT be honored, because the storage MAY allow for modifying files' metadata (in particular grants or other arbitrary metadata) without enforcing eventual locks.

As such, the spec changed from `MUST` to `SHOULD`, with an additional `MAY` clause.

This only affects the comments, the protobuf payloads did not change.

### DIFF
--- a/cs3/storage/provider/v1beta1/provider_api.proto
+++ b/cs3/storage/provider/v1beta1/provider_api.proto
@@ -205,7 +205,9 @@ message AddGrantRequest {
   Grant grant = 3;
   // OPTIONAL.
   // A lock_id: should the reference be locked, the stored
-  // lock_id MUST be equal to the given value.
+  // lock_id SHOULD be equal to the given value. However,
+  // storage implementations MAY allow for setting grants
+  // even if the lock does not match.
   string lock_id = 4;
 }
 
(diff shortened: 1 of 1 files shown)


glpatcern:
@butonic this was motivated by the fact that EOS allows to alter xattrs no matter the lock state, and I think it makes sense: e.g. I can (further) share a file even if it's being edited by someone. How did you implement the decomposedfs in Reva edge?

cs3/storage/provider/v1beta1/provider_api.proto
```diff
@@ -840,7 +846,9 @@ message UpdateGrantRequest {
   Grant grant = 3;
   // OPTIONAL.
   // A lock_id: should the reference be locked, the stored
-  // lock_id MUST be equal to the given value.
+  // lock_id SHOULD be equal to the given value. However,
+  // storage implementations MAY allow for setting grants
```
glpatcern, line 850:
```suggestion
  // storage implementations MAY allow for updating grants
```

cs3/storage/provider/v1beta1/provider_api.proto
```diff
@@ -786,7 +790,9 @@ message RemoveGrantRequest {
...
```
glpatcern, line 794:
```suggestion
  // storage implementations MAY allow for unsetting grants
```

PR #228, opened 2024-03-22
Enhancement: added ListExistingReceivedShares to gateway API

glpatcern:
This API is to be consumed by:
- the OCS (HTTP) service
- the LibreGraph API
- (in the future) the Jupyterlab integration

The idea is to incorporate in the gateway the current logic implemented (and duplicated) in other layers, where after `ListReceivedShares` one has to `Stat` each of them to validate it exists and get all necessary metadata to send the response (in the form of LibreGraph or OCS). This logic is to be implemented by the Gateway instead.

### DIFF
--- a/cs3/gateway/v1beta1/gateway_api.proto
+++ b/cs3/gateway/v1beta1/gateway_api.proto
@@ -35,6 +35,7 @@ import "cs3/permissions/v1beta1/permissions_api.proto";
 import "cs3/preferences/v1beta1/preferences_api.proto";
 import "cs3/rpc/v1beta1/status.proto";
 import "cs3/sharing/collaboration/v1beta1/collaboration_api.proto";
+import "cs3/sharing/collaboration/v1beta1/resources.proto";
 import "cs3/sharing/link/v1beta1/link_api.proto";
 import "cs3/sharing/ocm/v1beta1/ocm_api.proto";
 import "cs3/storage/provider/v1beta1/provider_api.proto";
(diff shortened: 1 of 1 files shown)


glpatcern:
@labkode @diocas : having discussed with @gmgigi96 I think it makes sense to have this in the Spaces implementation before going live with it, as the LibreGraph API does not cache the list of shares (that is currently implemented by the OCS layer only). It shouldn't be a big development but rather a code refactoring, where we move the code within the Gateway - the LibreGraph part is very simple, the OCS part a bit less but provided we agree on this API, I can look into it (I may need some help on the OCS part from @labkode).

cs3/gateway/v1beta1/gateway_api.proto
```diff
@@ -534,6 +538,26 @@ message ListAuthProvidersResponse {
...
+message ListExistingReceivedSharesResponse {
+  // REQUIRED.
+  // The response status.
+  cs3.rpc.v1beta1.Status status = 1;
+  // OPTIONAL.
+  // Opaque information.
+  cs3.types.v1beta1.Opaque opaque = 2;
+  // REQUIRED.
+  // The underlying shares as returned by the collaboration service.
+  repeated cs3.sharing.collaboration.v1beta1.ReceivedShare shares = 3;
```
labkode, line 550: I prefer having a new message encapsulating the ReceivedShare and the ResourceInfo to avoid the clients having to do the mapping themselves.

glpatcern: Makes sense, something like `repeated SharedResourceInfo shares = 3` with `SharedResourceInfo` being the encapsulating message (that should maybe live in `cs3.sharing.collaboration.v1beta1`?)

glpatcern: (Answering myself: no, this is a gateway resource)

labkode:
@glpatcern much better now! I don't understand the logic of moving it to the Spaces part unfortunately, the idea is to have an implementation in the gateway and adapt the upper clients, the caching can then be done at the gateway level so there is no need to modify OCS/Libregraph (just removing caching code).

glpatcern:
> I don't understand the logic of moving it to the Spaces part 

Bad wording, I rephrased my comment. Indeed the whole idea is to have the gateway encapsulate all the business logic and caching. This goes also along the lines of what we discussed with @butonic at CS3 a few weeks ago: OCS and LibreGraph would be a minimal translation layer (so code has to be removed) in front of the gateway.

Issue #235, opened 2024-10-10
Missing Options in cs3.storage.provider.v1beta1.ProviderAPI


marcdo7:
There are missing options in Proto file  _cs3.storage.provider.v1beta1.ProviderAPI_. 

I need to add these options, to get a working Java client

`option java_multiple_files = true;
option java_outer_classname = "ProviderApiProto";
option java_package = "com.cs3.storage.provider.v1beta1";`

labkode:
@marcdo7 please do

PR #236, opened 2024-10-11
Add options to build better java clients

marcdo7:
fixed https://github.com/cs3org/cs3apis/issues/235


labkode:
@jessegeens can you review?

glpatcern:
Hi @marcdo7, thanks for this addition, any specific reason why you need it on the storage provider API and not on the gateway API? A priori the latter is the "public" API that relays to all other APIs.

glpatcern:
I understand this is necessary (despite no further comments), so I'm going ahead and merge this.

glpatcern (approved):

PR #237, opened 2024-11-04
added a product property product_name to the app provider info

2403905:
We added a product property `product_name` to the `app provider info`, that could be used to handle product-specific differences. For example Collabora, OnlyOffice, Microsoft365 or MicrosoftOfficeOnline

### DIFF
--- a/cs3/app/registry/v1beta1/resources.proto
+++ b/cs3/app/registry/v1beta1/resources.proto
@@ -70,6 +70,11 @@ message ProviderInfo {
   // The action to be displayed to the user on the context menu.
   // By default this is "Open with".
   string action = 9;
+  // OPTIONAL.
+  // Specifies the kind of app provider.
+  // A product property that could be used to handle product-specific differences.
+  // For example: Collabora, OnlyOffice, Microsoft365 or MicrosoftOfficeOnline
+  string product_name = 10;
 }
 
 // Represents a mime type and its corresponding file extension.


kobergj (approved):

cs3/app/registry/v1beta1/resources.proto
```diff
@@ -70,6 +70,11 @@ message ProviderInfo {
   // The action to be displayed to the user on the context menu.
   // By default this is "Open with".
   string action = 9;
+  // OPTIONAL.
```
glpatcern, line 73: Even if already merged, I see this is de facto a breaking change, as you're going to use the `product_name` to identify the application "type": as such, this field should be `REQUIRED`.

Incidentally, Reva master does not take decisions based on the type (we just set all language headers for every app), and the CS3 wopiserver only has a few extra properties if the application is "Collabora". Nevertheless the new field makes perfect sense, to not risk inconsistencies with an admin-configurable (and user-display-oriented) field.

micbar: Ok. That makes sense. It means that we already had some hidden differences for the different products (collabora, codiMD and so on). Now we are making that official. IMHO we can keep it optional to avoid the breaking change on Reva master.

glpatcern: Well, I prefer to break Reva master now :) (those comments are anyhow advisory - unfortunately - meaning that the compilation goes through). At least to keep the API consistent, where I expect the HTTP app provider endpoint and the WOPI `/iop/open` one to change as well.

PR #241, opened 2025-06-26
Add expiration to UpdateShareRequest UpdateField

jessegeens:
This PR adds the `expiration` value to the `UpdateField` enum of the `UpdateShareRequest`, to allow expiration times to to be updated


glpatcern (approved):

glpatcern:
@micbar @butonic maybe this is one of the opaques you are using - otherwise how can OpenCloud pass the updated expiration time from libregraph to CS3API?

PR #246, opened 2025-08-11
Add thumbnail and readme IDs directly to space definition + add support for updating space quota and name

jessegeens:
The space definition now contains fields for the inodes of special files, specifically of the readme and thumbnail of a space


glpatcern (approved):
As discussed, this is to match the Libregraph API in full, even if some update actions may be disabled for administrative reasons (e.g. changing quota)

PR #247, opened 2025-09-02
Add support for filtering in FindUsers request

jessegeens:
This PR adds support to a `Filter` type in the `FindUsers` request, to filter for specific user types. The old `filter` parameter has been renamed to the more accurate `query`.

### DIFF
--- a/cs3/identity/user/v1beta1/user_api.proto
+++ b/cs3/identity/user/v1beta1/user_api.proto
@@ -138,9 +138,12 @@ message FindUsersRequest {
   // OPTIONAL.
   // Opaque information.
   cs3.types.v1beta1.Opaque opaque = 1;
-  // REQUIRED. TODO(labkode): create proper filters for most common searches.
-  // The filter to apply.
-  string filter = 2;
+  // REQUIRED. 
+  // The search query to apply.
+  string query = 4;
+  // OPTIONAL.
+  // Optional filters on users, such as filtering for a specific UserType.
+  repeated Filter filter = 5;
   // OPTIONAL.
   // Whether to skip fetching user groups along with the user object.
   bool skip_fetching_user_groups = 3;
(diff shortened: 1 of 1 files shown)


diocas (approved):

glpatcern:
Are we sure we want to introduce a breaking change?

jessegeens:
@butonic What do you think? If you want we can revert this change and keep the original `filter` name; otherwise you guys might get a breaking change in the API (rename `filter` -> `query`)

rhafer:
> Hi @rhafer, creating a new Request is also a breaking change, in the sense that compilation breaks when a new Request/Response is introduced but not implemented.

Actually (talking about the go implementation here) if the userprovider would properly embed the `UnimplementedUserAPIServer` struct in it's `service` struct. It should not result in compile errors.  That was invented exactly for this case AFAIK. But yeah currently we're not using that.

>  What I thought, at the expense of clarity, was to introduce the new Filter with a different name, such as `filterBy`.
> 
> Yet, many other CS3 APIs have `filter` with type `Filter`, and this was clearly overlooked at the beginning.

Agreed.

> So the breaking change does make sense after all.

Well, I'd just preferred if this would have been done in a none breaking manner. Additionally the implementation of FindUser() is getting somewhat clumsy IMO.
E.g. what is the expected behavior if both, `Query` and `Filter` are present? What if none is there?

rhafer:
> @rhafer Answering to my question, `query` is actually used as a generic query in several places. The logic is that `Filter` applies as post-processing after the query is performed: if no filter, return all results from `query` - see e.g. https://github.com/cs3org/reva/pull/5291/files#diff-d254163de2082008c31583d083be7d0564eb87bf795d036cc4d90b652e7ba2ac

But can't those cases where Query it still used be turned into a Filter?  I find the behavior of still having Query required and use Filter just as an additional "post Query" filter really un-intuitive. 

> Now, `query` is the former `REQUIRED` filter, so the behavior when none is present was and remains to return a `Bad Request`.
> 
> BTW a similar approach should be implemented for the Group filters, that would need a similar breaking change. Shall we bundle them and get the API fixed for good? I'd even have another breaking PR concerning OCM in the same vein!

I'd be still in favor of introducing a new Request Type for that. The 2nd best option would IMO be to get rid of the Query Parameter completely and only use Filter (maybe turn Query it into a  specific FilterType). But I really don't like to have both Query and Filter in one Request.

PR #248, opened 2025-09-03
Added support for federated groups + added comments

glpatcern:
This is in preparation for future "federated" groups, in particular looking after OCM-based federations.

Also this prepares a change similar to #247, which would be a breaking change for uniformity, where group filters are called `query`. @micbar I guess this specific PR is OK, and without further comments we will go ahead also for that.

### DIFF
--- a/cs3/identity/group/v1beta1/resources.proto
+++ b/cs3/identity/group/v1beta1/resources.proto
@@ -40,16 +40,45 @@ message GroupId {
   // the unique identifier for the group in the scope of
   // the identity provider.
   string opaque_id = 2;
+  // OPTIONAL.
+  // The type of group.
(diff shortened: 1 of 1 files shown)


cs3/identity/group/v1beta1/resources.proto
```diff
@@ -40,16 +40,45 @@ message GroupId {
   // the unique identifier for the group in the scope of
   // the identity provider.
   string opaque_id = 2;
+  // OPTIONAL.
+  // The type of group.
+  GroupType type = 9;
```
jessegeens, line 45: Is there a reason for jumping to `9` already?

jessegeens (approved):

PR #250, opened 2025-09-08
Add permissions to StorageSpace definition

jessegeens:


glpatcern (approved):
I think it's fine to keep this `OPTIONAL`, as with a bunch of other parameters that are not required for backwards compatibility. At some stage we have to review the "backwards compatibility" and decide the minimum set of parameters that make Spaces work.

PR #253, opened 2025-09-10
Introduced the Incoming OCM API. Fixes #150.

### LINKED ISSUE #150: How do `ocm/core` and `sharing/collaboration` relate?
While investigating https://github.com/cs3org/reva/issues/2148 I ran into a question:

Why is there on the one hand a [`cs3/ocm/core`](https://github.com/cs3org/go-cs3apis/tree/416ed6cf8b00baa3b05dcfa6c08ebfe6557f020c/cs3/ocm/core) package, with:
  * [`CreateOCMShareRequest`](https://github.com/cs3org/go-cs3apis/blob/416ed6c/cs3/ocm/core/v1beta1/ocm_core_api.pb.go#L31)

And on the other hand a [`cs3/sharing/collaboration`](https://github.com/cs3org/go-cs3apis/blob/416ed6cf8b00baa3b05dcfa6c08ebfe6557f020c/cs3/sharing/collaboration) package, with:
  * [`CreateShareRequest`](https://github.com/cs3org/go-cs3apis/blob/416ed6cf8b00baa3b05dcfa6c08ebfe6557f020c/cs3/sharing/collaboration/v1beta1/collaboration_api.pb.go#L66)

I think that `ocm/core` is used by the sender and `sharing/collaboration` is used by the receiver of a share, but if that's really the explanation then we should probably rename them to something like `ocm/sender` and `ocm/receiver`, respectively?

glpatcern:
This is a breaking change as a new API is introduced, but it serves as a one-on-one replacement of the so-called (and mis-named) "Core OCM" API, which is now deprecated and will be removed in a later change.

### DIFF
--- a/cs3/ocm/incoming/v1beta1/ocm_incoming_api.proto
+++ b/cs3/ocm/incoming/v1beta1/ocm_incoming_api.proto
@@ -18,7 +18,7 @@
 
 syntax = "proto3";
 
-package cs3.ocm.core.v1beta1;
+package cs3.ocm.incoming.v1beta1;
 
 import "cs3/identity/user/v1beta1/resources.proto";
 import "cs3/rpc/v1beta1/status.proto";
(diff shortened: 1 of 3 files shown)


glpatcern:
OK, merging now this change as we're also extending the possible OCM protocols and we want to do this with the new endpoints.

PR #261, opened 2026-04-01
Extend UserId

jessegeens:
Allows to store information specific to a certain identity provider and only consumed by this idp.
We would like to use this in the CERNBox reva-plugins to store some additional info about external users


glpatcern (approved):
Actually, as this PR only concerns the `UserId`, it is already complete. I'll open a separate PR for the `User`'s status.

rhafer (approved):

PR #265, opened 2026-04-14
Add new filter type for Spaceroots to share requests

rhafer:
This adds a new Filter type for the ListShares request. To be able to only List shares for SpaceRoots (i.e. Space memberships) and to be able to exclude SpaceRoot shares from the result.

### DIFF
--- a/cs3/sharing/collaboration/v1beta1/resources.proto
+++ b/cs3/sharing/collaboration/v1beta1/resources.proto
@@ -194,6 +194,10 @@ message Filter {
     TYPE_SPACE_ID = 7;
     TYPE_STATE = 8;
     TYPE_GRANTEE = 9;
+    // Filter shares by whether the shared resource is a space root.
+    // When the term `space_root` is true, only shares on space roots are returned.
+    // When false, shares on space roots are excluded.
+    TYPE_SPACE_ROOT = 10;
   }
   // REQUIRED.
   Type type = 2;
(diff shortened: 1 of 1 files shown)


micbar (approved):
@glpatcern Fine from POV

glpatcern (commented):
Looks ok for the semantic, but it seems now we have `ListStorageSpaces` and this filter that return the same thing in different formats. Or am I missing something?

rhafer:
> Looks ok for the semantic, but it seems now we have `ListStorageSpaces` and this filter that return the same thing in different formats. Or am I missing something?

There is some overlap, yes. But IMO the `ListShares` provides a more convenient interface. For `ListStorageSpaces` our implementation is currently putting quite some stuff into the `Opaque` which are much easier to consume when using `ListShares`. Also things like "List all Space that I have share with somebody else" is easier to achieve via the `ListShares` call than via `ListStorageSpaces`.

As a side note: Up to now our implementation didn't manage space memberships via the shareprovider at all (while we're using the gateway's `CreateShare` call to add a member to a space, the gateway is currently "re-routing" that call to the storage provider) our plan is too clean that up, and this PR is basically part of that effort.

`ListStorageSpaces` is still useful. E.g. for adminstrative task (e.g. to get quota and other space metadata).

glpatcern:
OK fair enough, it still seems weird to have two APIs to do the same thing, I guess `ListStorageSpaces` needs some changes for you to avoid using the opaques (we use it without opaques). At the same time the filter here is fine, I'll merge it.

PR #266, opened 2026-04-22
appprovider: add forced_view_mode_reason to response

glpatcern:


jessegeens (approved):

glpatcern:
@rhafer @micbar this is optional of course. The use case is in case a file is already opened/locked, we may downgrade in WOPI the access token to read-only, and need to communicate to Reva (and ultimately to the client) the reason why the permissions were downgraded.

PR #267, opened 2026-04-23
Replace favorites with a more generic labels API

jessegeens:
As a follow-up to https://github.com/cs3org/cs3apis/pull/260.

We would like to have a more generic version of the favorites API, where we can attach arbitrary labels to resources. These lables are user-scoped just like favorites. Migrating from AddFavorite to AddLabel would just be replacing the method name, and adding `Label: "favorite"`.


glpatcern:
@aduffeck @micbar @butonic what about this approach, instead of the `AddFavorite` you added in #260 ?

glpatcern (approved):

aduffeck:
That change should work for us. lgtm :+1:

glpatcern:
Thanks @aduffeck ! Let's go then

PR #268, opened 2026-04-28
Labels API

jessegeens:
Following up on https://github.com/cs3org/cs3apis/pull/267, we realized that we are also missing a way to `List` specific labels (like you would do `ListFavorites`). And since we made the API generic, we would also need a way to query the unique labels that a user has created. 

To make this a bit cleaner, we propose to add a Labels API. This one contains the two methods that were at the moment in the StorageProvider (`AddLabel`, `RemoveLabel`), as well as two new methods for listing unique labels and listing resources for a given label.

We store these in the database, so for us it makes sense to have them in a separate API from the storage provider. I think for others, if you do not need these, you can make them return unimplemented, and implement the API in the StorageProvider as you do now. That means that you would only need to add two stubs. 

What do you think @micbar @aduffeck @butonic ? Does this work for you?

### DIFF
--- a/cs3/gateway/v1beta1/gateway_api.proto
+++ b/cs3/gateway/v1beta1/gateway_api.proto
@@ -35,6 +35,7 @@ import "cs3/ocm/invite/v1beta1/invite_api.proto";
 import "cs3/ocm/provider/v1beta1/provider_api.proto";
 import "cs3/permissions/v1beta1/permissions_api.proto";
 import "cs3/preferences/v1beta1/preferences_api.proto";
+import "cs3/labels/v1beta1/label_api.proto";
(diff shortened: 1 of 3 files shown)


glpatcern (approved):
This change entails just a(nother) rename of APIs and payloads for the recently introduced `AddLabel` and `RemoveLabel` methods, that's why we didn't go for deprecation etc. (as we didn't when renaming the `SetFavorite` methods).

The extra methods are optional and not needed in a context of just dealing with favorites.

jessegeens:
@micbar @aduffeck @butonic do you approve? :slightly_smiling_face:

aduffeck (approved):
Sorry for the late reply, looks good to me.

PR #271, opened 2026-05-20
OCM: added an optional destination on ReceivedShare

glpatcern:
We are adding an optional field to `ocm.ReceivedShare`, for a receiving system when dealing with an OCM share that implies a data transfer: in that case, the flow is for the receiving user to "accept" the share and select a destination path. The extra field would be populated with the given path, so that the backend can execute the data transfer.


rawe0 (approved):
Looks good, it's just an addition.

jessegeens (approved):

PR #274, opened 2026-06-10
OCM: implement new webapp protocol

glpatcern:
In this PR we introduce support for the new payload of OCM `webapp` shares, as discussed and agreed in https://github.com/cs3org/OCM-API/pull/368.

We're also adding an optional `hidden` field to the ocm.ReceivedShare message, similar to `ReceivedShare.hidden`.

This is a breaking change, which we agreed within the OCM collaboration as the only working implementation of webapps is/was CERNBox, with the PoC we have done a few years ago. For this reason I didn't leave the dropped fields as `[deprecated]`. We plan to adapt our implementation to the new payload in the coming weeks.

In addition, the `permissions` for both the `webdav` and `webapp` OCM protocols were reviewed: we now use the `ResourcePermissions` message as opposed to a custom OCM message that did not bring any added value. The corresponding field for the `webdav` protocol is marked as `[deprecated]`.

### DIFF
--- a/cs3/sharing/ocm/v1beta1/resources.proto
+++ b/cs3/sharing/ocm/v1beta1/resources.proto
@@ -90,7 +90,8 @@ message Share {
   cs3.sharing.ocm.v1beta1.RecipientType recipient_type = 15;
 }
 
-// The permissions for a share.
+// The permissions for an OCM share.
+// Deprecated: reshare is redundant with grant permissions.
 message SharePermissions {
   storage.provider.v1beta1.ResourcePermissions permissions = 1;
   bool reshare = 2;
@@ -241,17 +242,6 @@ message ShareReference {
   }
 }
 
-// A share grant specifies the share permissions
-// for a grantee.
-message ShareGrant {
-  // REQUIRED.
-  // The grantee of the grant.
-  storage.provider.v1beta1.Grantee grantee = 1;
-  // REQUIRED.
-  // The share permissions for the grant.
-  SharePermissions permissions = 2;
-}
-
 // The protocol which is used to get access to a remote OCM resource.
 message Protocol {
   // REQUIRED.
@@ -276,9 +266,8 @@ message WebDAVProtocol {
   // REQUIRED.
   // Secret used to access the resource.
   string shared_secret = 1;
-  // REQUIRED.
-  // Permissions of the shared resource.
-  cs3.sharing.ocm.v1beta1.SharePermissions permissions = 2;
+  // Permissions of the shared resource. Deprecated: use share_permissions instead.
+  cs3.sharing.ocm.v1beta1.SharePermissions permissions = 2 [deprecated = true];
   // REQUIRED.
   // URI used to access the resource.
   string uri = 3;
@@ -288,6 +277,9 @@ message WebDAVProtocol {
   // OPTIONAL.
   // Access types for the share, e.g. remote or data transfer. Defaults to remote.
   repeated cs3.sharing.ocm.v1beta1.AccessType access_types = 5;
+  // REQUIRED.
+  // Permissions of the shared resource.
+  storage.provider.v1beta1.ResourcePermissions share_permissions = 6;
 }
 
 // Defines the options for the Webapp protocol, mapping
--- a/cs3/sharing/ocm/v1beta1/resources.proto
+++ b/cs3/sharing/ocm/v1beta1/resources.proto
@@ -168,6 +168,9 @@ message ReceivedShare {
   // Destination path of the share, in case the shared resource type is EMBEDDED
   // or the access type is DATATX.
   string destination = 18;
+  // OPTIONAL.
+  // Flag to hide the share, defaults to false.
+  bool hidden = 19;
 }
 
 // The state of the share.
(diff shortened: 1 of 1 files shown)


cs3/sharing/ocm/v1beta1/resources.proto
```diff
@@ -378,9 +399,27 @@ message WebDAVAccessMethod {
...
+  // A descriptive name of the remote app in the user interface.
+  string app_name = 6;
+  // OPTIONAL.
+  // A hint in the form of a media (MIME) type of the icon
+  // representing the app in the user interface.
+  string app_icon_hint = 7;
+  // OPTIONAL.
+  // An array of media (MIME) types that the application
+  // can handle.
+  repeated string media_types = 8;
```
glpatcern: We may not need these when we serve a share to a remote user, as we "know" them from the configuration. Let's confirm if we can drop them.

MahdiBaghbani:
nice 😄

cs3/sharing/ocm/v1beta1/resources.proto
glpatcern: (Dropped)

cs3/sharing/ocm/v1beta1/resources.proto
```diff
@@ -378,9 +399,19 @@ message WebDAVAccessMethod {
...
-  // The view mode for the share.
-  cs3.app.provider.v1beta1.ViewMode view_mode = 1;
+  // The permissions for the share.
+  cs3.sharing.ocm.v1beta1.SharePermissions permissions = 1;
+  // REQUIRED.
+  // The requirements for the share.
+  repeated string requirements = 2;
+  // REQUIRED.
+  // The targets that can be used to display the remote app.
+  repeated string targets = 3;
```
rawe0: Do we need to store the targets here? (In the access method)

My idea is that the remote system decides from the ones we offered via the share.

cs3/ocm/incoming/v1beta1/ocm_incoming_api.proto
```diff
@@ -157,6 +157,9 @@ message UpdateOCMIncomingShareRequest {
   // OPTIONAL.
   // Recipient type.
   cs3.sharing.ocm.v1beta1.RecipientType recipient_type = 8;
+  // OPTIONAL.
+  // Flag to hide the share, defaults to false.
+  bool hidden = 9;
```
rawe0: This should be in the ReceivedShare in sharing/ocm/v1beta1/*

cs3/sharing/ocm/v1beta1/resources.proto
rawe0: I assume this is the same as app_icon_hint and media_types as in it's useful for the remote system but for the local system we don't care as long as they use the one of the ones we have provided via the share.

glpatcern: Right, we can work them out anyway

cs3/sharing/ocm/v1beta1/resources.proto
```diff
@@ -267,36 +260,57 @@ message Protocol {
...
   // REQUIRED.
-  // Path-only URI to open the resource with a remote app.
-  string uri = 1;
+  // Secret used to access the resource.
+  string shared_secret = 1;
   // REQUIRED.
-  // View mode for the remote app.
-  cs3.app.provider.v1beta1.ViewMode view_mode = 2;
+  // Permissions of the shared resource.
+  storage.provider.v1beta1.ResourcePermissions share_permissions = 2;
```
glpatcern, line 293: A question I haven't answered yet: how are we going to map the `view-only` OCM permission in our CS3 `ResourcePermissions`?

rawe0: As discussed I currently do this (in a loop so it's additive).

```
        switch p {
        case "view":
            perms.Permissions.GetPath = true
            perms.Permissions.ListContainer = true
            perms.Permissions.Stat = true
        case "read":
            perms.Permissions.GetPath = true
            perms.Permissions.ListContainer = true
            perms.Permissions.Stat = true
            perms.Permissions.InitiateFileDownload = true
        case "write":
            perms.Permissions.InitiateFileUpload = true
        case "share":
            perms.Reshare = true
        }
```

rawe0: However, I think this is an implementation detail for Reva?

glpatcern: Well, here we're saying that _all_ implementations have to rely on the `storage.provider.v1beta1.ResourcePermissions` values.

The above might work, so let's go ahead for now. The concern is eventually the app needs to download the file in order to present it in `view-only` mode to the user, despite the user got permissions = `[Stat | ListContainer | GetPath]`. The "download" operation is done on the backend, so here we're saying that all implementations have to sort it out, e.g. with a special user on the backend (which is what we do in our Reva implementation).

rawe0 (approved):
Can be merged from my perspective, I'm happy with this.

glpatcern:
Given that this is backwards-compatible for OCM `webdav`, and that we are reimplementing the wiring of the OCM `webapp` in CS3 Reva, I'll go ahead and merge this.

PR #275, opened 2026-06-14
feat(gateway): add SetImmutable/UnsetImmutable RPCs

flash7777:
## Summary

Follow-up to #272. Exposes `SetImmutable` and `UnsetImmutable` on the Gateway API so that higher-level clients (like the OpenCloud Graph API) can call them through the gateway without needing direct StorageProvider access.

## Changes

- `cs3/gateway/v1beta1/gateway_api.proto`: add `SetImmutable` and `UnsetImmutable` RPCs using the existing request/response types from `cs3.storage.provider.v1beta1`

## Context

#272 added `SetImmutable`/`UnsetImmutable` to the StorageProvider API. The Reva gateway already proxies these calls (server-side), but the gateway proto client interface doesn't expose them yet. This blocks the OpenCloud Graph API from using these RPCs through the standard gateway client.

## Test plan

- [ ] Proto compiles
- [ ] Generated Go code includes new gateway client methods


glpatcern (changes requested):
Right, the external API is always the Gateway, I missed that the endpoints had to be replicated here.

Yet, can you please move the definitions below L212, as these are StorageProvider-related endpoints and don't stand on their own?

glpatcern (approved):

PR #276, opened 2026-07-10
Add events API

jessegeens:
Hello,

We would like to propose an event-based notifications API in the CS3APIs. 

We are reworking our implementation of notifications in Reva and we think this is a nice way to handle it. The idea is that there are certain "events" in reva that a user might want to be notified of (e.g. a share is created, file is uploaded in a folder, etc). There are many ways these can be handled (email, push notification, etc.). Users might also have preferences for how often they want to be contacted, and for certain event types you also want to accumulate some events (e.g. a user uploading 100 files). 

We decided to make it more generic, so that events can also be used to trigger other actions. With this API, it is completely up to the handler to handle all of this; all you set is an `event_type` and some data and the implementation then chooses what to do with it (we will use a system where you can define rules for event types in the config).

What do you think (@aduffeck @micbar @butonic)?

### DIFF
--- a/cs3/gateway/v1beta1/gateway_api.proto
+++ b/cs3/gateway/v1beta1/gateway_api.proto
@@ -81,6 +81,15 @@ service GatewayAPI {
   // WhoAmI returns the information for a user.
   rpc WhoAmI(WhoAmIRequest) returns (WhoAmIResponse);
   // *****************************************************************/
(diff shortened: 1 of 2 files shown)


glpatcern (approved):
I think we can go ahead with this non-breaking change (let alone the need to define stubs if not implemented).

PR #278, opened 2026-08-11
add creation_time to lock definition

jessegeens:
@aduffeck @micbar @butonic any objections? :)


diocas (approved):
It's an optional field. Might be useful to display that info for users, in case a file has been locked for a long time, might lead to the realization that it was forgotten.

PR #279, opened 2026-08-27
add roles to space definition

jessegeens:
Hi guys,

We would like to add "role definitions" on a space. We need a way to, from a space definition, figure out who has which permissions in the space. Right now we used the `permissions` field which was scoped to the owner on request. But this has some issues: for some things we need the actual groups. On top of that, with the heterogeneous spaces we will likely also have more fine-grained roles. 

What do you think (@aduffeck @butonic @micbar ) ?

### DIFF
--- a/cs3/storage/provider/v1beta1/resources.proto
+++ b/cs3/storage/provider/v1beta1/resources.proto
@@ -545,9 +545,33 @@ message StorageSpace {
   // OPTIONAL.
   // An opaque ID of an optional readme file of the space (usually the inode)
   string readme_id = 14;
-  // OPTIONAL.
+
+  // DEPRECATED.
   // The set of permissions for the storage space effective for the authenticated user.
-  ResourcePermissions permission_set = 15;
+  // ResourcePermissions permission_set = 15;
+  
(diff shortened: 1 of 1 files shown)


glpatcern (approved):
The removing of the current, simpler `permission_set` is motivated by the fact that we added it in #250 and we assume it was not used by you. Of course if it is we can keep it - or keep it as `[deprecated]`.

cs3/storage/provider/v1beta1/resources.proto
```diff
@@ -546,8 +546,27 @@ message StorageSpace {
   // An opaque ID of an optional readme file of the space (usually the inode)
   string readme_id = 14;
   // OPTIONAL.
-  // The set of permissions for the storage space effective for the authenticated user.
-  ResourcePermissions permission_set = 15;
+  // A set of roles that members of the space can have,
+  // each role being assigned to a set of grantees and
+  // yielding a set of permissions.
+  repeated SpaceRole roles = 16;
```
glpatcern, line 557: Let's decide if this is `16` with `15` deprecated...

